PR #156, opened 2021-01-18
Disable HPA in prod

AntoineAugusti:
Temporarily disable HPA in production by hardcoding min/max to expected and sensible values: 2 admin, 4 API, 6 Celery workers


jimleroyer (approved):
Good thinking on resizing the min and max to be the same. That will help me see if the 502 load balancer alarm is triggered or not. Thank you.

PR #207, opened 2021-02-01
feat: update CONTACT_EMAIL env in staging

AntoineAugusti:
Same PR than in https://github.com/cds-snc/notification-manifests/pull/193 but in staging

PR #218, opened 2021-02-03
Bump production to 0.24.0

AntoineAugusti:
Releasing https://github.com/cds-snc/notification-manifests/pull/207 to production, added the new env variable in the production file as well. It's safe to release in advance, what's important is that this PR goes out before https://github.com/cds-snc/notification-api/pull/1222 because we'll need this new env variable in production.

0.23.0 is unrelated, it was https://github.com/cds-snc/notification-manifests/pull/207


jimleroyer (approved):
:shipit:

PR #222, opened 2021-02-04
Update number of Celery pods in production

AntoineAugusti:
## What are you changing?
- [x] Changing Kubernetes configuration

We have been recently paged during business hours because we don't clear up the email queue fast enough. The last occurrence was [yesterday](https://gcdigital.slack.com/archives/CNWA63606/p1612378640091100).

We processed ~7000 emails/5 minutes with 6 Celery pods, this amounts to 23 emails sent per second. Our limit is at 100 emails sent per second so I'm picking up a 2.5x increase (max replicas from 6 to 15) to clear the queue faster while being under the SES rate limit.

HPA for Celery pods in production is safe compared to HPA with admin/API pods so I'd like to give this a go.

### DIFF
--- a/env/production/replica_count.yaml
+++ b/env/production/replica_count.yaml
@@ -26,7 +26,7 @@ metadata:
   name:  celery
   namespace: notification-canada-ca
 spec:
-  replicas: 6
+  replicas: 3
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -62,5 +62,5 @@ metadata:
   name: celery-hpa
   namespace: notification-canada-ca
 spec:
-  minReplicas: 6
-  maxReplicas: 6
+  minReplicas: 3
+  maxReplicas: 15


env/production/replica_count.yaml
```diff
@@ -62,5 +62,5 @@ metadata:
   name: celery-hpa
   namespace: notification-canada-ca
 spec:
-  minReplicas: 6
-  maxReplicas: 6
+  minReplicas: 3
+  maxReplicas: 15
```
AntoineAugusti, line 66: We've got HPA with Celery pods in staging and this is not causing us problems even when doing load tests

https://github.com/cds-snc/notification-manifests/blob/9c16ca2993c2a22047061b2204868cd7f297c27f/env/staging/replica_count.yaml#L49-L56

jimleroyer (approved):

PR #251, opened 2021-02-17
Reduce pool size in production to 10

AntoineAugusti:
## What are you changing?

Reduce `SQLALCHEMY_POOL_SIZE` from 15 to 10 in production


jimleroyer (approved):
Decrypted the file locally and confirm pool size is set at 10. 🚢

PR #257, opened 2021-02-19
Set CPU and memory for ipv4 deployment

AntoineAugusti:
## What are you changing?

Allocate a minimum of CPU and memory to the IP geolookup deployment.

Set to double what it was currently using when I looked at the data

```
kubectl top pod ipv4-8578d99fc-ncblp 
NAME                   CPU(cores)   MEMORY(bytes)   
ipv4-8578d99fc-ncblp   26m          41Mi
```

### DIFF
--- a/base/ipv4-deployment.yaml
+++ b/base/ipv4-deployment.yaml
@@ -22,6 +22,10 @@ spec:
       containers:
         - name: ipv4
           image: public.ecr.aws/cds-snc/notify-ipv4-geolocate-webservice:latest
+          resources:
+            requests:
+              cpu: "50m"
+              memory: "100Mi"
           ports:
             - containerPort: 8080
           readinessProbe:


base/ipv4-deployment.yaml
```diff
@@ -22,6 +22,10 @@ spec:
       containers:
         - name: ipv4
           image: public.ecr.aws/cds-snc/notify-ipv4-geolocate-webservice:latest
+          resources:
+            requests:
+              cpu: "50m"
+              memory: "100Mi"
```
RenDavidDimen, line 28: Just curious where these values come from, are these simply norms?

RenDavidDimen (approved):

base/ipv4-deployment.yaml
AntoineAugusti: You can read this section to know what are the units on Kubernetes https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes

About the values, I explained in the PR description how I picked those 👍

PR #274, opened 2021-03-03
feat: add env DOCUMENTATION_DOMAIN in admin and API

AntoineAugusti:
## What are you changing?

Stop relying on [the fallback value in `config.py`](https://github.com/cds-snc/notification-admin/blob/f947e0d2053b372d2321670dcf7d0260287dba8d/app/config.py#L32), the fallback mechanism is going away soon. Introduces the env variable on the API side as well, it will be needed in https://github.com/cds-snc/notification-api/pull/1241


QuanNguyen-CDS (approved):
Good!

PR #284, opened 2021-03-04
Reduce max number of Celery pods in production

AntoineAugusti:
## What are you changing?

Lower the number of Celery pods while we work on some other performance issues related to CPU/database connections.


dorianjp (approved):
carefully lgtm

PR #326, opened 2021-03-17
feat: improve document download deployments

AntoineAugusti:
## What are you changing?

Now that we are going to start sending more attachments, make sure that those deployments are more robust:

- `RollingUpdate` strategy for deployments, for 0 downtime deployments
- Define CPU + memory `requests` and `limits` to give pods a minimum amount of resources to run
- Add `readinessProbe`s to make sure new pods are not serving trafic when the Docker container is not ready yet

Trello card: https://trello.com/c/OSYe4SV2/386-make-document-download-k8s-deployments-more-robust

### DIFF
--- a/base/document-download-api-deployment.yaml
+++ b/base/document-download-api-deployment.yaml
@@ -3,7 +3,7 @@ kind: Deployment
 metadata:
   labels:
     app: document-download-api
-  name:  document-download-api
+  name: document-download-api
   namespace: notification-canada-ca
 spec:
   replicas: 1
@@ -11,6 +11,11 @@ spec:
   selector:
     matchLabels:
       app: document-download-api
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
   template:
     metadata:
       labels:
@@ -51,7 +56,24 @@ spec:
               value: '$(NEW_RELIC_MONITOR_MODE)'
             - name: NOTIFY_ENVIRONMENT
               value: '$(ENVIRONMENT)'
-          resources: {}
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "400Mi"
+            limits:
+              cpu: "400m"
+              memory: "800Mi"
+          ports:
+            - containerPort: 7000
+          readinessProbe:
+            httpGet:
+              path: /_status
+              port: 7000
+            initialDelaySeconds: 1
+            periodSeconds: 3
+            timeoutSeconds: 1
+            successThreshold: 3
+            failureThreshold: 10
       dnsPolicy: ClusterFirst
       restartPolicy: Always
       schedulerName: default-scheduler
--- a/base/document-download-frontend-deployment.yaml
+++ b/base/document-download-frontend-deployment.yaml
@@ -3,7 +3,7 @@ kind: Deployment
 metadata:
   labels:
     app: document-download-frontend
-  name:  document-download-frontend
+  name: document-download-frontend
   namespace: notification-canada-ca
 spec:
   replicas: 1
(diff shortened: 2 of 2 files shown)


base/document-download-api-deployment.yaml
```diff
@@ -3,14 +3,19 @@ kind: Deployment
...
   replicas: 1
   revisionHistoryLimit: 5
   selector:
     matchLabels:
       app: document-download-api
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
```
AntoineAugusti, line 18: Those are the defaults, but now it's explicit. https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#rolling-update-deployment 

Because we're running 1 replica, `maxSurge` will be 1 and `maxUnavailable` will be 0

base/document-download-api-deployment.yaml
```diff
@@ -51,7 +56,24 @@ spec:
...
             - name: NOTIFY_ENVIRONMENT
               value: '$(ENVIRONMENT)'
-          resources: {}
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "400Mi"
+            limits:
+              cpu: "400m"
+              memory: "800Mi"
```
AntoineAugusti, line 65: Ran `kubectl top pods` + compared those values to admin and API pods

base/document-download-api-deployment.yaml
```diff
@@ -51,7 +56,24 @@ spec:
...
+          ports:
+            - containerPort: 7000
+          readinessProbe:
+            httpGet:
+              path: /_status
```
AntoineAugusti, line 70: Does not call other services https://github.com/cds-snc/notification-document-download-api/blob/master/app/healthcheck.py

base/document-download-frontend-deployment.yaml
```diff
@@ -54,7 +54,24 @@ spec:
...
             - name: NOTIFY_ENVIRONMENT
               value: '$(ENVIRONMENT)'
-          resources: {}
+          resources:
+            requests:
+              cpu: "50m"
+              memory: "400Mi"
+            limits:
+              cpu: "400m"
+              memory: "800Mi"
```
AntoineAugusti, line 63: Ran `kubectl top pods` + compared those values to admin and API pods

base/document-download-frontend-deployment.yaml
```diff
@@ -54,7 +54,24 @@ spec:
...
+          ports:
+            - containerPort: 7001
+          readinessProbe:
+            httpGet:
+              path: /_status
```
AntoineAugusti, line 68: Does not call other services https://github.com/cds-snc/notification-document-download-frontend/blob/3a650d03d3f64c789d3bcebcb66624ceb0555523/app/main/views/index.py#L11-L13

base/document-download-api-deployment.yaml
```diff
@@ -51,7 +56,24 @@ spec:
...
-          resources: {}
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "400Mi"
+            limits:
+              cpu: "400m"
+              memory: "800Mi"
+          ports:
+            - containerPort: 7000
```
AntoineAugusti, line 67: Is not really needed https://github.com/kubernetes/kubernetes/issues/4332#issuecomment-448900396 but maps to the same values https://github.com/cds-snc/notification-manifests/blob/2772a8b4ea096c770538a7858b45ae01242425c7/base/document-download-api-service.yaml#L11-L12

base/document-download-frontend-deployment.yaml
```diff
@@ -54,7 +54,24 @@ spec:
...
-          resources: {}
+          resources:
+            requests:
+              cpu: "50m"
+              memory: "400Mi"
+            limits:
+              cpu: "400m"
+              memory: "800Mi"
+          ports:
+            - containerPort: 7001
```
AntoineAugusti, line 65: Same value than the service https://github.com/cds-snc/notification-manifests/blob/2772a8b4ea096c770538a7858b45ae01242425c7/base/document-download-frontend-service.yaml#L10-L12

dorianjp (approved):
great walkthrough on the changes

PR #350, opened 2021-03-31
fix: adding ASSET_DOMAIN to the api container

QuanNguyen-CDS:
## What are you changing?

Adding ASSET_DOMAIN which is not defined within notification-api container. 
ASSET_DOMAIN within the code defaults to the production address

- [x] API
- [ ] Admin
- [ ] Document API
- [ ] Document UI

### DIFF
--- a/base/api-deployment.yaml
+++ b/base/api-deployment.yaml
@@ -112,6 +112,8 @@ spec:
               value: 'http://api.notification-canada-ca.svc.cluster.local:6011'
             - name: ASSET_UPLOAD_BUCKET_NAME
               value: '$(ASSET_UPLOAD_BUCKET_NAME)'
+            - name: ASSET_DOMAIN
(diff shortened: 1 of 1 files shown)


AntoineAugusti (approved):

AntoineAugusti:
Made sure it's defined in `.env`

QuanNguyen-CDS:
> Made sure it's defined in `.env`

Yes, that's the first thing I did was to decrypt the `.env.enc.aws` and was pleasantly surprised `ASSET_DOMAIN` was present

AntoineAugusti:
You can repeat this change on the init container. It’s not really needed, it’s only use to do DB migrations. We keep those in sync

QuanNguyen-CDS:
> You can repeat this change on the init container. It’s not really needed, it’s only used to do DB migrations. We keep those in sync

good to know, let me update the PR then.

PR #358, opened 2021-04-06
feat: added autoscaling to document api pods

QuanNguyen-CDS:
## What are you changing?

feat: Giving document-api the ability to autoscale within staging, note the cpu limit is 50 milliCPU under the  container limit

- [ ] API
- [ ] Admin
- [x] Document API
- [ ] Document UI


env/production/replica_count.yaml
```diff
@@ -64,3 +64,12 @@ metadata:
 spec:
```
AntoineAugusti, line 64: The production file will need to be added after bumping the tag

AntoineAugusti (changes requested):

env/production/replica_count.yaml
QuanNguyen-CDS: thanks for the review and the catch - updated to remove

AntoineAugusti (approved):

AntoineAugusti:
Could you submit a prod bump when you get a chance?

PR #494, opened 2021-06-04
Brought in configuration recent changes in environment

jimleroyer:
## What are you changing?

* Reduced pool size from 15 to 10 per worker instances in the staging environment. 
* Applied production changes of replicas into staging. 

Because we are stress testing our applications system, we need a staging environment that looks more like production so that we can properly optimize and fine tune performance.

- [ ] API
- [ ] Admin
- [ ] Document API
- [ ] Document UI


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,6 +8,7 @@
 ## If you are releasing a new version of notify, what components are you updating
 - [ ] API
 - [ ] Admin
+- [ ] Documentation
```
jimleroyer, line 11: I've observed this was missing from our components list.

env/staging/replica_count.yaml
```diff
@@ -20,6 +20,16 @@ spec:
 ---
 apiVersion: apps/v1
 kind: Deployment
+metadata:
+  labels:
+    app: celery
+  name:  celery
+  namespace: notification-canada-ca
+spec:
+  replicas: 3
```
jimleroyer, line 29: I added a section for `celery` as the production file has one but not staging. I assume this is for the desired count of instances.

env/staging/replica_count.yaml
```diff
@@ -35,15 +45,15 @@ metadata:
   namespace: notification-canada-ca
 spec:
   minReplicas: 2
-  maxReplicas: 4
+  maxReplicas: 2
```
jimleroyer, line 48: I downgraded this one to replicate production, as staging had more defined than production which I found curious. Tell me if that was intended and I will revert back.

.github/PULL_REQUEST_TEMPLATE.md
AntoineAugusti: Thanks for adding it, I've been wanting to do that for a while and I've been lazy.

env/staging/replica_count.yaml
AntoineAugusti: We disabled this in production because we didn't test enough HPA for admin/API so we just put 4 API pods and 2 admin pods in production, which is enough by experience

AntoineAugusti (approved):
Decoded the encrypted env file and it's fine!

PR #574, opened 2021-07-12
Added env var to disable api for services

jimleroyer:
## What are you changing?
- [ ] Releasing a new version of Notify
- [X] Changing kubernetes configuration

## Provide some background on the changes

Following [this PR](https://github.com/cds-snc/notification-admin/pull/1071/files) which disables access to API related pages in the admin component, we need to add an environment variable to control which services will be restricted. Changes in this PR prepare the way to test for the staging environment. 

## If you are releasing a new version of notify, what components are you updating
- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document API
- [ ] Document UI

## Checklist if releasing new version:
- [ ] I made sure that both API and Admin changes are present in Notify
- [ ] I have checked if the docker images I am referencing exist - ex: https://gallery.ecr.aws/v6b8u5o6/notify-admin

## Checklist if making changes to Kubernetes:
- [ ] I know how to get kubectl credentials in case it catches on fire

PR #584, opened 2021-07-14
revert: adding the env variable API_DISABLED_SERVICE_IDS

jimleroyer:
## What are you changing?

This reverts the feature for *disable the API permission for some services* which was introduced [by this PR](https://github.com/cds-snc/notification-manifests/pull/574).

After speaking to Bryan, it seems the solution as not good enough for RCMP and their security team moved on from this issue. As the feature is currently broken, we can get rid of the feature altogether. We can bring back the idea and this code if we have the same feature request but we'd have to fix the logic contained in these previous changes.

I also added some `.devcontainer` repository for VSCode. This will bootstrap an remote container environment under a minute. This is quick to install proper tools for this project for first timer or non-regular installation. For example, this is useful to me to generate the targets in the `Makefile` under an Alpine Linux command line. 

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document API
- [ ] Document UI


jimleroyer:
[Modify a single RCMP service to not allow them to create API keys for any reason](https://trello.com/c/GrV6fssa/574-modify-a-single-rcmp-service-to-not-allow-them-to-create-api-keys-for-any-reason)

Moro-Code (approved):
Looks good 👍🏽

PR #586, opened 2021-07-15
fix: adding variable to application container

Moro-Code:
## What are you changing?

I made an oopsie and realised I added the variable to the init container but not the application one. So I fixed that

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document API
- [ ] Document UI

### DIFF
--- a/base/api-deployment.yaml
+++ b/base/api-deployment.yaml
@@ -204,6 +204,8 @@ spec:
               value: 'https://api.getbase.com'
             - name: ZENDESK_SELL_API_KEY
               value: '$(ZENDESK_SELL_API_KEY)'
+            - name: FF_NOTIFICATION_CELERY_PERSISTENCE
+              value: '$(FF_NOTIFICATION_CELERY_PERSISTENCE)'
           resources:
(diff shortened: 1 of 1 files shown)


base/api-deployment.yaml
```diff
@@ -204,6 +204,8 @@ spec:
               value: 'https://api.getbase.com'
             - name: ZENDESK_SELL_API_KEY
               value: '$(ZENDESK_SELL_API_KEY)'
+            - name: FF_NOTIFICATION_CELERY_PERSISTENCE
+              value: '$(FF_NOTIFICATION_CELERY_PERSISTENCE)'
```
jimleroyer, line 208: Damn, I did noticed that this entry wasn't in there when I worked on restricting the API for RCMP. I thought you figured something out and that was not necessary.

jimleroyer: And in fact, I think it might not be the only one missing IIRC. We should double-check for other entries.

jimleroyer (approved):

PR #602, opened 2021-07-22
feat: Detach the database check from the k8s readiness probe

jimleroyer:
## What are you changing?

Tweaking the readiness probe for both admin and API components to remove the database check. For now, removing it will prevent these to go down if the database is slowed to a point that the components can't reply in a timely manner. The API component especially should not go down and stop processing because the database is in `autovacuum` processing.

More details in the [ADR for the database bottleneck](https://github.com/cds-snc/notification-adr/pull/1).

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document API
- [ ] Document UI

### DIFF
--- a/base/admin-deployment.yaml
+++ b/base/admin-deployment.yaml
@@ -105,7 +105,7 @@ spec:
             - containerPort: 6012
           readinessProbe:
             httpGet:
-              path: /_status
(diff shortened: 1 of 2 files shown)


Moro-Code (approved):
Looks good to me !

PR #635, opened 2021-08-09
Revert "lower max celery workers (#634)"

jimleroyer:
## What are you changing?

This reverts commit d8f25e0069addd11b64373126f158eb054d10321.

Part of Omar's testing in staging involved seeing if reducing the number of celery workers would reduce database reads and transitively reduce the read loads experienced in staging. We should eventually revert this test to re-align with what we have in production, unless we decide to change the latter as well.


sastels (approved):
:shipit:

PR #637, opened 2021-08-10
Target production RDS proxies

jimleroyer:
## What are you changing?

Changed both `POSTGRES_SQL` and `SQLALCHEMY_DATABASE_READER_URI` environment variables to target the RDS proxies in production.

[Part of ADR #1](https://github.com/cds-snc/notification-adr/blob/main/records/2021-07-16.scalability.database-bottleneck.md) decisions is to use RDS proxies to have more connection reuse with a global pool manager.

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document API
- [ ] Document UI


env/staging/replica_count.yaml
```diff
@@ -63,7 +63,7 @@ metadata:
   namespace: notification-canada-ca
 spec:
   minReplicas: 3
-  maxReplicas: 5
+  maxReplicas: 10
```
jimleroyer, line 66: Oh, that slipped in, but it should be increased back again to `10` to match the value in production.

Moro-Code (approved):

sastels (approved):
:shipit:

PR #795, opened 2021-11-19
Removed redirection to /dev/null for kubectl kustomize command.

jimleroyer:
## What are you changing?

Removed redirection to /dev/null for kubectl customize command.

We could not quickly assess the reason for the failed build following the `kubectl kustomize`  command from that build. We would like to have all the output for now even if that is a long output. Better than silence.

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document API
- [ ] Document UI

### DIFF
--- a/.github/workflows/syntax_check.yaml
+++ b/.github/workflows/syntax_check.yaml
@@ -14,4 +14,4 @@ jobs:
 
       - name: Run manifest build
         run: |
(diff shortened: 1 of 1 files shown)


mohamed-cds (approved):
LGTM 🚀 ✅ . This should make future debugging easier if this fails again

PR #810, opened 2021-12-07
Update read me with API Lambda update details

iokpala:
Within we have added a sentience to direct readers to the API Lambda section of the terraform docs


sastels (approved):
🚀

PR #833, opened 2022-01-12
Add config to switch on Batch insertion

jzbahrai:
## What are you changing?
In order to test https://github.com/cds-snc/notification-api/pull/1426

We switch on the FF_BATCH_INSERTION: true in staging.


sastels (changes requested):
Will also need to change `env/staging/kustomization.yaml` and (I think?) `base/celery-deployment.yaml` to get the variable into kustomize and then into the pods. (I think it's just the celery pods? what about celery-sms? I don't think api is involved?)

sastels (approved):
yep, I think this'll do it now! Looking forward to load testing.

PR #850, opened 2022-01-18
Add config for batch insertion chunk size

jzbahrai:
## What are you changing?
Adding configs for chunk size. 
This config is used for batch size in the bulk insertion function in notification-api.
We want it to be configurable so that we can optimize the chunk size for the best functionality. Making it a configurable number will help with testing that.


iokpala (approved):
LGTM

jimleroyer (approved):
Locally ran the `make diff-staging` and `make diff-production` commands and it looked good!

PR #871, opened 2022-01-31
fix: use correct api key

mohamed-cds:
Zendesk api key was accidently left in a reverted state when the api keys were being corrected. This PR updates the zendesk sell production api key with the correct value.


jimleroyer (approved):
Made a diff locally, saw the value was different from previous one. LGTM

PR #879, opened 2022-02-03
Reset the API HPA to previous values, 4min/8max scalability

jimleroyer:
## What are you changing?

We went through an incident this week where two clients hit the API hard for a short amount of time, possibly the root cause of bringing the API Kubernetes pod instances into an out of memory state. This caused restarts and a few messages to be dropped (around 20-25 observed in AWS CloudWatch / Logs Insights).

image

image

- [x] API
- [x] Admin
- [ ] Documentation
- [ ] Document download API
- [ ] Document download frontend

    - [ ] [document download frontend](https://gallery.ecr.aws/v6b8u5o6/notify-document-download-frontend)

## After merging this PR

### DIFF
--- a/env/production/replica_count.yaml
+++ b/env/production/replica_count.yaml
@@ -45,7 +45,7 @@ metadata:
   namespace: notification-canada-ca
 spec:
   minReplicas: 2
-  maxReplicas: 2
+  maxReplicas: 3
 ---
 apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
@@ -54,7 +54,7 @@ metadata:
   namespace: notification-canada-ca
 spec:
   minReplicas: 4
-  maxReplicas: 4
+  maxReplicas: 8
 ---
 apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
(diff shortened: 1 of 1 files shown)


env/production/replica_count.yaml
```diff
@@ -72,4 +72,4 @@ metadata:
   namespace: notification-canada-ca
 spec:
   minReplicas: 1
-  maxReplicas: 1
+  maxReplicas: 2
```
sastels, line 75: We probably don't need to ever have more than one `document-download-api` pod, but it doesn't hurt to have another available.

sastels (approved):
My only concern would be if it's possible to overload something else if the # pods get boosted (ie this would have killed the database back when api directly wrote new notifications). I can't think of anything likely though (ie we shouldn't get thousands of hits / minute of someone wanting updated status on a notification or something).

jimleroyer:
We'll definitely need to keep an eye on applied pressure at key points in the data pipeline: API, SQS queues, workers, the database and Redis.

env/production/replica_count.yaml
jimleroyer: I increased it as I saw some errors in the past where the cause might have been insufficient resources. But I didn't investigate long enough to be certain of it.

PR #889, opened 2022-02-11
feat: pin replica count patches to a specific tagged release

mohamed-cds:
This fixes the issue related to replica counts unexpectedly being deployed to production. This PR will pin production configurations to a tagged release using a VERSION file.

The new workflow will be:
1. Make changes to production deployment, approve & merge
2. Tag the release
3. Release PR that bumps the VERSION file to the newly created tag in step 2.


maxneuvians:
> and could potential introduce errors in the future related to forgetting to bump both git file references

Could this be mitigated with automation or a check to make sure they are in sync?

jimleroyer:
Reading the shared document about the design, I might be reading this wrong, but I have the impression the folks are not against people using a template system but rather, they want to stay away from it for kustomize, reasons they provide which makes sense. There is this paragraph that would be telling to me:

> Other tools (sed, jinja, erb, envsubst, kafka, helm, ksonnet, etc.) provide varying degrees of unstructured editting and/or embedded languages, and can be used instead of, or in a pipe with, kustomize. If you want to go all-in on configuration as a language, consider [cue](https://cuelang.org/).

So we could start with something simple to store a few variables. As long as the changes are traceable and versioned, and the consumed file by kubernetes are yaml in/out for consistency, we should be OK.

I am fine with the approach where we make sure the values are in check too, if deemed more appropriate.

mohamed-cds:
Reviewing your comments @jimleroyer  & @maxneuvians  + an earlier discussion with @patheard, it seems that all `yaml` files will need a specific git tag in-order to be pinned to a version. Currently all items under [resources](https://github.com/cds-snc/notification-manifests/blob/2a8307b83ae014dd8dadd8242f2ca17b048298e0/env/production/kustomization.yaml#L3-L9) are locally referenced, so we'd run into the immediate prod deployment as seen with the patched `replica_count.yaml`.

I've pushed a [new change](https://github.com/cds-snc/notification-manifests/pull/889/commits/158e09aee34eda940fe3f27731d22d687fea8858) that moves this repo towards using a VERSION file similar to other projects. The new workflow would be:

1. Make changes to production deployment, approve & merge
2. Tag the release
3. Release PR that bumps the VERSION file to the newly created tag in step 2.

The main advantage is that the versioning is now centrally located and we can say with certainty what is in production vs individually tagging yaml files in the `kustomization.yaml`. This also eliminates the potential of using multiple tags with different values.

.github/workflows/merge_to_main_production.yaml
```diff
@@ -24,6 +24,16 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Set Prod Version
+        run: |
+          TARGET_VERSION=`cat VERSION`
```
patheard, line 29: Minor tweak - we'll need to update the workflow trigger so that it runs on `VERSION` file changes:
```yaml
  push:
    branches:
      - main
    paths:
      - ".github/workflows/merge_to_main_production.yaml"
      - "VERSION"
```

env/production/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 bases:
-  - github.com/cds-snc/notification-manifests//base?ref=v0.44.2
+  - ../../base
```
jimleroyer, line 2: Working and simple, love it. 👍

.github/workflows/merge_to_main_production.yaml
mohamed-cds: Good catch, updated! Thanks 👍🏾

jimleroyer (approved):
LGTM!

PR #952, opened 2022-02-26
feat: workflow to test environment variables

patheard:
# Summary
Add a GitHub workflow that checks if the Example, Staging and
Production env variable names are in sync.  This is to catch cases
where environment variable changes are not properly made to
all environments.

# Related
* cds-snc/notification-planning-core#10

# What are you changing?
- [x] GitHub workflow

### DIFF
--- a/.github/workflows/test_env_vars.yaml
+++ b/.github/workflows/test_env_vars.yaml
@@ -22,6 +22,7 @@ jobs:
       - name: Get example env var keys
         run: |
           ENV_KEYS_EXAMPLE="$(make env-keys-example)"
+          echo "::add-mask::$ENV_KEYS_EXAMPLE"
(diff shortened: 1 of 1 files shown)


mohamed-cds (changes requested):
This is really nice! One suggestion is to check that every entry in the file contains a `=`. This will prevent accidentally dumping secrets in the event of a malformed entry in a `.env` that doesn't contain a `=`.

Something like `cat test.txt| grep -v '=' | wc -l` > 0 ; abort with malformed and don't print diff

patheard:
That's a really good idea.  I've added in `grep="="` so that only environment variables with an equal sign will be split and compared.

I've also updated the action output so that only the `diff` is printed to further limit the chance of any accidental secret leaking.

jimleroyer (approved):
👏

PR #969, opened 2022-03-07
feat: add test manifest build action

patheard:
# Summary
Add an action that tests a Kustomize manifest build to check
if it can build successfully and there are no environment
variables that are not used or replaced.

# What are you changing?
- [x] GitHub workflows

# Related

### DIFF
--- a/.github/actions/test-manifest/action.yaml
+++ b/.github/actions/test-manifest/action.yaml
@@ -4,15 +4,18 @@ inputs:
   build-target:
     description: Name of the manifest build target
     required: true
+  echo-manifest:
+    description: Echo the built manifest to the logs
+    default: "false"
+    required: false
 
 runs:
   using: "composite"
   steps:
     - name: Test manifest build
       shell: bash
-      run: |
-        exec 5>&1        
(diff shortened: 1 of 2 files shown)


jimleroyer:
I wonder if making this required is practical, if there are situations where we might be out of sync anyway for some valid reason. We can try this out and see how it goes.

patheard:
> I wonder if making this required is practical, if there are situations where we might be out of sync anyway for some valid reason. We can try this out and see how it goes.

It's definitely possible that having it required will become a problem.  

Here are the errors it's currently reporting:

### Staging
```sh
Run ./.github/actions/test-manifest
Run EXIT=0
          value: $(SENSITIVE_SERVICES)
❌ Undefined environment variable found
```

### Production
```sh
Run ./.github/actions/test-manifest
Run EXIT=0
          value: $(GC_ARTICLES_API_AUTH_USERNAME)
          value: $(GC_ARTICLES_API_AUTH_PASSWORD)
❌ Undefined environment variable found
```

In both cases it's because the variable isn't defined in the environments `kustomization.yaml` file.

PR #973, opened 2022-03-07
fix: missing environment variables in kustomize build

patheard:
# Summary
Add environment variables to the kustomize build that are defined
but not replaced.  The result of this change is that instead of
the manifests containing values of `$(VAR_NAME)`, they will
have the value set in the environment's `.env` file.

- **Staging:** Add `SENSITIVE_SERVICES` variable.
- **Production:** Add `GC_ARTICLES_API_AUTH_*` variables.

In both cases, the current `.env` value is blank `""`. 

## What are you changing?

# Related
* #969 

### DIFF
--- a/env/production/kustomization.yaml
+++ b/env/production/kustomization.yaml
@@ -428,3 +428,17 @@ vars:
       apiVersion: v1
     fieldref:
       fieldpath: data.FF_REDIS_BATCH_SAVING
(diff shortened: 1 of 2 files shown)


jzbahrai (approved):

jimleroyer (approved):
Should be good. The GCArticles env vars have no fallback on configuration and it's only using staging now.

PR #978, opened 2022-03-08
feat: automatically release to production when the auto-pr bot release is merged

mohamed-cds:
This PR introduces automation that will trigger the deployment of new releases automatically when the Notify PR bot auto-generated release is merged.

This is done by tracking merged pull requests containing the title [AUTO-PR]. PR's containing the title of [MANIFEST] will also trigger the same process, so that changes to the Kubernetes config can be automatically deployed using the tag system.

This is done by using a new GitHub workflow to bump the `VERSION` file in the root directory, and to update the main branch with this change. Manual intervention will no longer be required when releasing updates to the apps managed by this repo. To achieve this, this new GitHub worklow will disable branch protection, push the update to the version file, and then re-enable branch protection.

[Demo repo that was used to test this workflow](https://github.com/cds-snc/test_autopr/actions/runs/1958375304)

All changes that require authentication will be using the Notify PR bot token which has access to make changes to this repo.

Workflow
1. `notification-pr-bot` generates a new release candidate PR
2. PR approved and merged
3. New workflow kicks off

```mermaid
flowchart TD
A[PR merged] --> B["PR title contains [AUTO-PR] or [MANIFEST]"]
B -- Yes --> C[Tag the release] --> E[Bump the version file] --> F[Turn off branch protection] --> G[Commit and push updated version file to main] --> H[Turn on branch protection] --> I[New production release starts deploying]
B -- No --> D[No change to production deployment]

```
image

Related: cds-snc/notification-planning#429

### DIFF
--- a/.github/workflows/autopr_merged.yaml
+++ b/.github/workflows/autopr_merged.yaml
@@ -23,10 +23,8 @@ jobs:
           github_token: ${{ env.GITHUB_TOKEN }}
           release_branches: main
 
-      - name: Checkout newly merged pr
+      - name: Checkout main branch
         uses: actions/checkout@v2
-        with:
-          ref: 'refs/pull/${{ github.event.number }}/merge'
 
       - name: Get current Prod Version
         run: |
--- a/.github/workflows/autopr_merged.yaml
+++ b/.github/workflows/autopr_merged.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   tag-release:
     runs-on: ubuntu-latest
-    if: ${{ github.event.pull_request.base.ref == 'main' && github.event.pull_request.merged && (contains(github.event.pull_request.title, '[AUTO-PR]') || contains(github.event.pull_request.title, '[MANIFEST]')) }}
+    if: github.event.pull_request.base.ref == 'main' && github.event.pull_request.merged && (contains(github.event.pull_request.title, '[AUTO-PR]') || contains(github.event.pull_request.title, '[MANIFEST]'))
     
     steps:
       - name: Obtain a Notify PR Bot GitHub App Installation Access Token
@@ -31,6 +31,18 @@ jobs:
           CURRENT_VERSION=`cat VERSION`
           echo "CURRENT_VERSION=$CURRENT_VERSION" >> $GITHUB_ENV
 
+      - name: Update version file
+        env:
+          NEW_VERSION: ${{ steps.tag_version.outputs.new_tag }}
+          CURRENT_VERSION: ${{ env.CURRENT_VERSION }}
+        run: |
+          echo $NEW_VERSION > VERSION
+
+      - name: setup git config
+        run: |
+          git config user.name "Notify PR Bot"
+          git config user.email "<>"
+
       - name: Branch protection OFF
         uses: octokit/request-action@v2.x
         with:
@@ -46,18 +58,6 @@ jobs:
             null 
         env:
           GITHUB_TOKEN: ${{ env.GITHUB_TOKEN }}
-
-      - name: Update version file
-        env:
-          NEW_VERSION: ${{ steps.tag_version.outputs.new_tag }}
-          CURRENT_VERSION: ${{ env.CURRENT_VERSION }}
-        run: |
-          echo $NEW_VERSION > VERSION
-
-      - name: setup git config
-        run: |
-          git config user.name "Notify PR Bot"
-          git config user.email "<>"
           
       - name: commit
         env:
--- a/.github/workflows/autopr_merged.yaml
+++ b/.github/workflows/autopr_merged.yaml
@@ -10,11 +10,17 @@ jobs:
     if: ${{ github.event.pull_request.base.ref == 'main' && github.event.pull_request.merged && (contains(github.event.pull_request.title, '[AUTO-PR]') || contains(github.event.pull_request.title, '[MANIFEST]')) }}
     
     steps:
+      - name: Obtain a Notify PR Bot GitHub App Installation Access Token
+        run: |
+          TOKEN="$(npx obtain-github-app-installation-access-token@1.1.0 ci ${{ secrets.GH_APP_CREDENTIALS_TOKEN }})"
+          echo "::add-mask::$TOKEN"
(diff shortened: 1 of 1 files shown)


.github/workflows/autopr_merged.yaml
```diff
@@ -0,0 +1,90 @@
...
+    if: ${{ github.event.pull_request.base.ref == 'main' && github.event.pull_request.merged && (contains(github.event.pull_request.title, '[AUTO-PR]') || contains(github.event.pull_request.title, '[MANIFEST]')) }}
+    
+    steps:
+      - name: Obtain a Notify PR Bot GitHub App Installation Access Token
+        run: |
+          TOKEN="$(npx obtain-github-app-installation-access-token@1.1.0 ci ${{ secrets.GH_APP_CREDENTIALS_TOKEN }})"
+          echo "::add-mask::$TOKEN"
+          echo "GITHUB_TOKEN=$TOKEN" >> $GITHUB_ENV
+
+      - name: Bump version and push tag
```
jimleroyer, line 19: Will this always bump on the patch, minor or major version? how does that work?

.github/workflows/autopr_merged.yaml
```diff
@@ -0,0 +1,90 @@
+name: "Tag and release to production"
+
+on: 
+  pull_request:
+    types: [closed]
+
+jobs:
+  tag-release:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.base.ref == 'main' && github.event.pull_request.merged && (contains(github.event.pull_request.title, '[AUTO-PR]') || contains(github.event.pull_request.title, '[MANIFEST]')) }}
```
patheard, line 10: nit: you can get rid of the wrapping `${{ }}` here

.github/workflows/autopr_merged.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Checkout main branch
+        uses: actions/checkout@v2
+
+      - name: Get current Prod Version
+        run: |
+          CURRENT_VERSION=`cat VERSION`
+          echo "CURRENT_VERSION=$CURRENT_VERSION" >> $GITHUB_ENV
+
+      - name: Branch protection OFF
```
jimleroyer, line 46: Does disabling the branch protection only apply within the execution scope of this workflow action or to all of the repository?

.github/workflows/autopr_merged.yaml
```diff
@@ -0,0 +1,90 @@
...
+        env:
+          NEW_VERSION: ${{ steps.tag_version.outputs.new_tag }}
+          CURRENT_VERSION: ${{ env.CURRENT_VERSION }}
+        run: |
+          echo $NEW_VERSION > VERSION
+
+      - name: setup git config
+        run: |
+          git config user.name "Notify PR Bot"
+          git config user.email "<>"
```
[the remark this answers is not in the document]
mohamed-cds: makes total sense.

.github/workflows/autopr_merged.yaml
mohamed-cds: done and done

.github/workflows/autopr_merged.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Branch protection OFF
+        uses: octokit/request-action@v2.x
+        with:
+          route: PUT /repos/:repository/branches/main/protection
+          repository: ${{ github.repository }}
+          required_status_checks: |
+            null
+          enforce_admins: |
+            null
```
patheard, line 54: Another question - if we just disabled `enforce_admins` and the bot performing this action had an `admin` token, it would lower the risk of branch protection being disabled since it would only impact `admins` if it didn't get re-enabled properly (or there was a race-condition).

On the other-hand, it would mean our bot was `admin` scoped.  What do you think?

mohamed-cds: let me try it in my test repo!

.github/workflows/autopr_merged.yaml
```diff
@@ -0,0 +1,90 @@
+name: "Tag and release to production"
+
+on: 
+  pull_request:
+    types: [closed]
```
jimleroyer, line 5: Interesting change of the flow. I was expecting we would rather have the VERSION merge be the latest manual step but yeah, we can keep it the way it is right now with the manual merge of the pr-bot changes and then automate the rest.

jimleroyer (commented):
Not blocking this PR but I feel this could require some documentation. In the [release management doc](https://docs.google.com/document/d/17kVWBN4OK_mSgTK7h_dTyC1kpjcY_UhXOn1FpolH-GM/edit) but also of the overall process with the provided diagram, potentially as an ADR. I am concerned that people from the future will need quick understanding of the release process at some point and such doc could help.

.github/workflows/autopr_merged.yaml
mohamed-cds: The action will parse the new commits since the last tag using the [semantic-release](https://github.com/semantic-release/semantic-release) conventions and decide automatically what it should do.

.github/workflows/autopr_merged.yaml
mohamed-cds: only applies to this workflow and would be `OFF` for a very short duration before being enabled again.

.github/workflows/autopr_merged.yaml
mohamed-cds: thinking about this again, i don't think giving the bot admin is a good idea just in-case the token somehow gets leaked. This would allow adding users, changing secrets, etc.

.github/workflows/autopr_merged.yaml
mohamed-cds: i figured it's more intuitive when someone merges the PR bot release that a release would happen and not have to worry about tagging and updating version files.

.github/workflows/autopr_merged.yaml
patheard: makes sense - I couldn't decide which was worse 😅

patheard (approved):
for science 🧪

PR #983, opened 2022-03-09
fix: tweak checkout step to not prefix

mohamed-cds:
Workflow tweak to no longer prefix with a `v` and to checkout the base branch

### DIFF
--- a/.github/workflows/autopr_merged.yaml
+++ b/.github/workflows/autopr_merged.yaml
@@ -25,6 +25,8 @@ jobs:
 
       - name: Checkout main branch
         uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/autopr_merged.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
 
       - name: Checkout main branch
         uses: actions/checkout@v2
+        with:
+          ref: $GITHUB_HEAD_REF
```
jimleroyer, line 29: 💥

jimleroyer (approved):
Let's try this out, for science like @patheard would say 🧑‍🔬 🧪

PR #987, opened 2022-03-10
fix: disable all branch protection for auto-PR

patheard:
# Summary
Disabling only the `enforce_admin` was not enough to allow the
Notify PR bot to commit the VERSION file change.

This also includes the VERSION file change that was skipped
because of the error, which will release #985 to Production.


sastels (approved):
This seems like black magic but let's see if it works.

jimleroyer (approved):

PR #1013, opened 2022-03-29
fix: switch to `git cat-file` to get previous .env

patheard:
# Summary
Update the `diff-staging` and `diff-production` Makefile targets to get
the previous version of the environment file from the repo history rather
than using curl.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,13 +18,13 @@ decrypt-staging-orig:
 
 decrypt-previous-staging:
 	@cd env/staging &&\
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,13 +18,13 @@ decrypt-staging-orig:
 
 decrypt-previous-staging:
 	@cd env/staging &&\
-	curl -L -o .previous.env.zip.enc.aws https://github.com/cds-snc/notification-manifests/blob/main/env/staging/.env.zip.enc.aws?raw=true > /dev/null 2>&1 &&\
+	git cat-file blob main:env/staging/.env.zip.enc.aws > .previous.env.zip.enc.aws &&\
```
jimleroyer, line 21: Just catching up on this, very nice change. 👍

PR #1088, opened 2022-05-06
fix: install pinned kubectl without Azure action

patheard:
# Summary
Update all workflows that use `kubectl` to use a pinned version
installed from the official release.

# Related
* #1086


jimleroyer (approved):
:shipit:

PR #1097, opened 2022-05-10
switch on priority lanes

jzbahrai:
➜  notification-manifests git:(task/switch-on-priority) ✗ make diff-staging
Archive:  .previous.env.zip
  inflating: .env
Archive:  .env.zip
  inflating: .env
38c38
< FF_PRIORITY_LANES=false
---
> FF_PRIORITY_LANES=true


patheard (approved):
Looks good!  

Also validated that production is still set to `FF_PRIORITY_LANES=false`

PR #1110, opened 2022-05-16
fix

jzbahrai:
switch on Priority lanes

```
➜  notification-manifests git:(task/on) ✗ make diff-staging
Archive:  .previous.env.zip
  inflating: .env
Archive:  .env.zip
  inflating: .env
38c38
< FF_PRIORITY_LANES=false
---
> FF_PRIORITY_LANES=true
```


sastels (approved):
third time's the charm? 😅

PR #1132, opened 2022-05-30
switch on priority queues

jzbahrai:
## What are you changing?
Switch on priority lanes

```
Archive:  .previous.env.zip
  inflating: .env
Archive:  .env.zip
  inflating: .env
38c38
< FF_PRIORITY_LANES=false
---
> FF_PRIORITY_LANES=true
```


jimleroyer (approved):

jimleroyer:
Smoke tests might have failed due to the currently running performance tests running in staging. We can retry to run the smoke test GH action once these are done.

PR #1211, opened 2022-08-30
add slack alert for failing github actions

sastels:
## What are you changing?

Add an alert to slack if the GitHub actions fail


sastels:
I believe the `testing_manifest` will never return since there's no changes to the "real" code in this PR.

jimleroyer (approved):
Ça l'air bon!

PR #1252, opened 2022-09-28
Try lifecycle change

jzbahrai:
## What are you changing?
Try lifecycle change before the container is stopped. This is related to 
https://github.com/cds-snc/notification-api/pull/1610


base/celery-deployment.yaml
```diff
@@ -106,6 +106,12 @@ spec:
               value: '$(NEW_RELIC_MONITOR_MODE)'
             - name: FF_CLOUDWATCH_METRICS_ENABLED
               value: 'True'
+          lifecycle:
+          preStop:
+            exec:
+              command: ["/bin/bash"]
```
patheard, line 112: Is `/bin/bash` in the API docker image?  If not, could you get away with using `/bin/sh`?

jzbahrai: it works when I checked it manually, I think it is okay.

jimleroyer (approved):
Let's try this! 🎆

PR #1423, opened 2023-01-11
Change make-diff to check again origin

jzbahrai:
We want to check diffs against what is in current staging and prod and not local stg/prd

When manifest changes, we need to check the change against origin, not against our local checked out version of main.

## Testing
I tested this by changing main (locally), making a new branch (locally) and running `make diff-staging`
It checks the result against local main. 
With my changes, I see the correct result


andrewleith (approved):
LGTM!

PR #1507, opened 2023-03-08
Fix: Celery resource usage and performance tuning

ben851:
Increasing the memory limit on celery pods to 1024Mi. Also configure horizontal auto scaler to scale out based on memory usage

## What are you changing?

Celery is running into OOM errors in production due to relatively low memory limits. Additionally, the horizontal pod autoscaler (HPA) is not currently configured to monitor memory usage. 

### DIFF
--- a/base/celery-hpa.yaml
+++ b/base/celery-hpa.yaml
@@ -17,3 +17,9 @@ spec:
       target:
         type: Utilization
         averageUtilization: 50
+  - type: Resource
+    resource:
+      name: memory
+      target:
+        type: Utilization
+        averageUtilization: 75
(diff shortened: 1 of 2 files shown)


base/celery-hpa.yaml
```diff
@@ -17,3 +17,9 @@ spec:
       target:
         type: Utilization
         averageUtilization: 50
+  - type: Resource
+    resource:
+      name: memory
+      target:
+        type: Utilization
+        averageUtilization: 75
```
jimleroyer, line 25: Very useful, thanks for this one.

jimleroyer (approved):
:shipit: (and let's test the OOM if possible in staging)

PR #1540, opened 2023-03-21
rolling over new relic api keys

ben851:
Rolled over new relic api keys. I made separate API Keys for prod and staging for better security.

## What are you changing?
- [ ] Updating encrypted environment zips with rolled over new relic API keys

This is part of the procedure for periodic password rollovers. 

## Testing
- [ ] We will need to deploy this to staging to verify that the API key is working as expected. Once completed, we can promote to production

## Cleanup
- [ ] Once promoted to production, we will have to revoke the old API Key.


sastels (approved):
LGTM!

PR #1544, opened 2023-03-22
reverting new relic api key and merging salesforce env vars

ben851:
## What are you changing?
- [ ] Reverting the encrypted env vars to use the old new relic license key while we debug why the lambda function does not properly use the new license key.

We are going through a rollover of api keys for new relic, which is working in all K8s deployments but there seems to be a problem with the way the lambda functions are reading the api keys. We need to revert this to allow other releases while we debug this problem


sastels (approved):
👍  verified with make diff-* that we are reverting the new relic keys back to what they were before.

PR #1550, opened 2023-03-23
Rolling over API key for new relic in production

ben851:
The API key for new relic in production will be rolled over to a new value that is separated from staging.

## What are you changing?
Production encrypted .env file

We are going through a secret rollover exercise, and this is the PR for rolling over new relic in production (already done in staging).

- [x] I made sure that the changes are as expected in [Notify staging](https://staging.notification.cdssandbox.xyz/)


sastels (approved):
verified locally that only changes the  prod`NEW_RELIC_LICENSE_KEY`

PR #1610, opened 2023-04-06
Prometheus cloudwatch

ben851:
Kube State Metrics and AWS Cloudwatch Prometheus Agent will be installed in the staging kubernetes cluster.

## What are you changing?
- [x] Adding new features to kubernetes

We had an incident action item of creating cloudwatch alarms when K8s resources were in a failed state. Cloudwatch does not natively ingest the required metrics for these alerts. I've added kube-state-metrics to pull the necessary metrics, and then installed and configured the aws cloudwatch prometheus agent to scrape KSM and forward the metrics to cloudwatch.

## After merging this PR


base/prometheus-cloudwatch/cwagent-iam.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: cwagent-prometheus-role
+rules:
+  - apiGroups: [""]
+    resources:
+    - nodes
+    - nodes/proxy
+    - services
+    - endpoints
+    - pods
+    verbs: ["get", "list", "watch"]
```
patheard, line 13: Minor nit, but could you use the same `verbs` list style that you had for the `ksm-iam.yaml`:
```suggestion
    verbs:
    - get
    - list
    - watch
```

env/production/cwagent-configmap.yaml
```diff
@@ -0,0 +1,71 @@
...
+    {
+      "agent": {
+        "region": "ca-central-1",
+        "debug": false
+      },
+      "logs": {
+        "metrics_collected": {
+          "prometheus": {
+            "cluster_name": "notification-canada-ca-production-eks-cluster",
+            "log_group_name": "/aws/containerinsights/notification-canada-ca-production-eks-cluster/prometheus",
```
patheard, line 18: I wish we didn't have to hardcode these values but I can't think of a way to get the value dynamically either.

base/prometheus-cloudwatch/cwagent-iam.yaml
ben851: I just copy/pasted this part of the config - I will refactor it to match our existing structure - consistency is good!

env/production/cwagent-configmap.yaml
patheard: I think you're right about the only option being `envsubst`.  Appreciate you digging into it more though.

patheard (approved):

PR #1679, opened 2023-05-03
Adding the github actions new relic deployment notification api key

ben851:
The production env file will now include the secret value for the new relic api key for deployment notifications. While this value is not referenced by kubernetes/kustomize, it is specific to this repo, so I've added it here until we get Vault going.

## What are you changing?
- Production .env

Rolling over Github Actions New Relic Deployment Notification API Key


ben851:
Closing PR because we're going to just keep it in last pass for now.

PR #1721, opened 2023-05-23
fix: Add `FF_BOUNCE_RATE_BACKEND` to yaml files

smcmurtry:

Release this first to staging with the `fix:` prefix.


andrewleith (approved):
LGTM - new config value is now everywhere the old one was except admin!

PR #1736, opened 2023-05-30
[MANIFEST] Scaling out prod documentation and doc download in prep for k8s upgrade

ben851:
Documentation and Document Download APIs will scale out to 2 pods minimum. This is in preparation for the K8s upgrade so that there will always be at least one pod running.

## What are you changing?

Required for ensuring no downtime while upgrading k8s to 1.23

- [X] Documentation
- [X] Document download API

- [X] I made sure that the changes are as expected in [Notify staging](https://staging.notification.cdssandbox.xyz/)

## After merging this PR

### DIFF
--- a/env/production/replica_count.yaml
+++ b/env/production/replica_count.yaml
@@ -36,7 +36,7 @@ metadata:
   name:  documentation
   namespace: notification-canada-ca
 spec:
-  replicas: 1
+  replicas: 2
 ---
 apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
@@ -71,5 +71,5 @@ metadata:
   name: document-download-api-hpa
   namespace: notification-canada-ca
 spec:
-  minReplicas: 1
+  minReplicas: 2
   maxReplicas: 4


env/production/replica_count.yaml
```diff
@@ -71,5 +71,5 @@ metadata:
   name: document-download-api-hpa
   namespace: notification-canada-ca
 spec:
-  minReplicas: 1
+  minReplicas: 2
```
jimleroyer, line 74: Is this a requirement of the new upgrade? Or we just want to make this more reliable?

jimleroyer (approved):

env/production/replica_count.yaml
ben851: When the upgrade happens, K8s will try and move over only one pod at a time. If there's only one pod running, it will terminate first and then move over, which will introduce downtime. With two running, when it terminates the first one, the other will handle the traffic until the new one comes up.

jimleroyer: ah yes, makes total sense, thanks!

jimleroyer: Although thinking about it, Kubernetes usually spins up a new pod before killing the old one. But I guess this is different for upgrading its own pod?

PR #1752, opened 2023-06-02
HPA api version migrated from beta to v2 for k8s 1.24

ben851:
The horizontal pod autoscalers for GC Notify will be converted to the v2 api spec away from the v2beta api spec which is no longer supported as of K8s v1.24

## What are you changing?

K8s v1.24 removes support for autoscaling/v2beta api. I have converted the HPAs for notify to the new v2 api format. Tested working in my scratch account

### DIFF
--- a/base/admin-hpa.yaml
+++ b/base/admin-hpa.yaml
@@ -4,6 +4,7 @@ metadata:
   name: admin-hpa
   namespace: notification-canada-ca
 spec:
+  minReplicas: 1
   maxReplicas: 2
   metrics:
   - resource:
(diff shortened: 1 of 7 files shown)


base/admin-hpa.yaml
```diff
@@ -1,19 +1,19 @@
...
-  - type: Resource
-    resource:
+  - resource:
       name: cpu
       target:
-        type: Utilization
         averageUtilization: 50
+        type: Utilization
+    type: Resource
+  minReplicas: 1
```
jimleroyer, line 15: Nitpick: can we have the minReplicas and maxReplicas side by side, near each others?

jimleroyer (approved):
Approved but had a small change request, if that can be done prior to merging.

base/admin-hpa.yaml
ben851: Yes good point - I ran this through a converter app and should have reorganized its output.

PR #1765, opened 2023-06-06
Fluentbit

ben851:
Staging will move from fluentd to fluentbit for log aggregation.

## What are you changing?

Fluentd breaks in v1.24 of kubernetes - since it is so fragile, I've moved us to fluent bit which seems to be better supported for EKS.


env/scratch/fluentbit.yaml
```diff
@@ -0,0 +1,375 @@
...
+  name: fluent-bit-config
+  namespace: amazon-cloudwatch
+  labels:
+    k8s-app: fluent-bit
+data:
+  fluent-bit.conf: |
+    [SERVICE]
+        Flush                     5
+        Grace                     30
+        Log_Level                 info
```
jimleroyer, line 63: This is for the logging level of fluentbit itself I assume?

env/scratch/fluentbit.yaml
```diff
@@ -0,0 +1,375 @@
...
+    [OUTPUT]
+        Name                cloudwatch
+        Match               application.*
+        region              ${AWS_REGION}
+        log_group_name      /aws/containerinsights/${CLUSTER_NAME}/application
+        log_stream_name     $(tag[4])
+        auto_create_group   true
+        extra_user_agent    container-insights
+
+  dataplane-log.conf: |
```
jimleroyer, line 139: Why is it called data plane?

jimleroyer (approved):
Looks good, as far as I can tell. 😅

env/scratch/fluentbit.yaml
ben851: Yeah the [SERVICE] section is the fluentbit config section

PR #1874, opened 2023-07-13
Reverting celery MLP fix and trying the skip long lines

ben851:
We will revert the fix for fluent bit celery MLP and attempt to resolve the dropped logs using the skip_long_lines On configuration globally. Also add file size limits to fluentbit

**Staging Only**

## What are you changing?

Follow up incident action for missing api and admin logs with fluentd in cloudwatch


jimleroyer (approved):
Let's break stuff. 🔨

PR #1882, opened 2023-07-20
Updating Document Download S3 Bucket Name

ben851:
- This will update the document download s3 bucket name to the new name set in notification-terraform

## What are you changing?

We shortened the name of document download bucket since it was nearing character limits. We need to update the dd-api to point to the new name

- [X] Document download API (environment var change)

## After merging this PR


whabanks (changes requested):
Once the conflict is resolved this looks good to go, checked out the diffs for both prod and staging.

PR #1888, opened 2023-07-25
Fluent bit staging alias

ben851:
Fluentbit in staging will now have aliases set for the different plugins, and debug log level is enabled

## What are you changing?

To aid in troubleshooting fluent-bit dropping logs.

## After merging this PR


sastels (approved):
sure why not!

env/staging/fluentbit.yaml
```diff
@@ -132,22 +137,24 @@ data:
...
         Name                cloudwatch
+        Alias               notify-app-logs-cloudwatch-output
         Match               application.*
         region              ${AWS_REGION}
         log_group_name      /aws/containerinsights/${CLUSTER_NAME}/application
         log_stream_name     $(tag[4])
         auto_create_group   true
         extra_user_agent    container-insights
-        Log_Level           info
+        Log_Level           debug
```
jimleroyer, line 151: I don't think we want debug set to go as a default and in production! Can we revert these?

PR #1895, opened 2023-07-26
Fluentbit mlp fix

ben851:
Fluentbit MLP regex settings will be updated in staging. Celery and App logs will be consolidated into one log stream, reducing memory usage. 

## What are you changing?

This is in response to the fluent-bit log flushing incident


env/scratch/fluentbit.yaml
```diff
@@ -306,8 +337,10 @@ data:
         flush_timeout 1000
         # rules |   state name  | regex pattern         | next state
         # ------|---------------|-----------------------------------
-        rule      "start_state"   "/^\[.*\].*/"                "cont"
-        rule      "cont"          "/^[^\[].*/"           "cont"
+        rule      "start_state"   "/^\b\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\b|^\[.*\]/"           "cont"
+        rule      "cont"          "/^[\s].*|^\b(Traceback|Exception)\b.*/"  "cont"
```
jimleroyer, line 341: That seems good enough, but I wonder if we had cases where multi line continuation was required and there were no whitespace characters in the beginning? 🤔 Let's test this thoroughly once it hits staging. Tell me and  I can help with the testing.

jimleroyer (approved):
LGTM

env/scratch/fluentbit.yaml
ben851: Yeah and that's where the problem occurred... if there's no whitespace, the regex picks up all new log entries... I tried doing a pattern "not" like IP address or date, but couldn't get it working. If this proves to be insufficient, maybe we can link up and try and debug my other pattern.

PR #1913, opened 2023-07-31
Celery HPA CPU only

ben851:
The HPA for celery in scratch and staging will set to scale based off of CPU usage only, since this process seems to be more CPU bound.

## What are you changing?

Part of celery performance tuning

## After merging this PR
- [x] I have verified that I can still log into [Notify production](https://notification.canada.ca)
- [x] I have verified that the smoke tests still pass on production
- [x] I have communicated the release in the #notify Slack channel.

### DIFF
--- a/base/celery-hpa.yaml
+++ b/base/celery-hpa.yaml
@@ -6,19 +6,6 @@ metadata:
 spec:
   minReplicas: 1
   maxReplicas: 2
-  metrics:
-  - resource:
-      name: cpu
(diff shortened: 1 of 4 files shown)


sastels (approved):
LGTM!
We'll have to remember to do this for prod as well if we decide it's a good idea

ben851:
Somethings up with the PR - I'm going to try reopening it.

PR #1916, opened 2023-07-31
Upping celery HPA max to 40

ben851:
Maximum number of celery workers will increase to 40 in staging only.

## What are you changing?

Re: Celery performance tuning


sastels (approved):
📈

PR #1931, opened 2023-08-03
Setting termination grace period to 60

ben851:
Setting termination grace period to 60 on all k8s deployments to allow for proper connection draining

## What are you changing?

This in support of doing node maintenance with k8s cordon and drain.


sastels (approved):
:shipit:

jimleroyer (approved):

PR #1932, opened 2023-08-03
Karpenter Code Only

ben851:
This PR will configure Karpenter in staging and production. Karpenter will provision only m5.large and r5.large spot instances. 

Additionally, celery and celery-sms are configured to run only on the karpenter spot instances.
Additionally Aditionally, the front apps are configured to only run on the on-demand instances (regular ones) so that they do not go down with the epehemerality (yeah that's a word) of the spot instances.

## What are you changing?

In support of celery performance tuning


jimleroyer:
So we're officially migrating to helm?

ben851:
> So we're officially migrating to helm?

Well the issue is that there is no base kustomize template for karpenter... so it was helm or I have to make one... I'm not against doing it but for this PoC I took the easy way.

ben851:
Note for the future - looks like there's a decent way to convert helm to kustomize:
https://github.com/Spazzy757/helm-to-kustomize

sastels:
Will the `celery-sms-send` pods on staging also be restricted to the spot instances?

ben851:
> Will the `celery-sms-send` pods on staging also be restricted to the spot instances?

That was the idea but it doesn't have to be this way

base/kustomization.yaml
```diff
@@ -34,3 +34,4 @@ resources:
   - kube-state-metrics/ksm-deployment.yaml
   - kube-state-metrics/ksm-service-account.yaml
   - kube-state-metrics/ksm-service.yaml
+
```
sastels, line 37: Some extra spaces or something here? not sure what the diff is

sastels (approved):
I think I have a rough understanding of what these files are doing! LGTM. What does @jimleroyer think?

PR #1937, opened 2023-08-08
Renaming replica_count.yaml ttuning celery HPA

ben851:
The celery HPA in staging and scratch will not scale more than 4 pods at a time with a 60 second refresh window. This is very conservative, but good for a baseline.  We will have to play with these numbers to find a good balance between speed and stability.

Also renamed replica_count.yaml to performance.yaml to better reflect what's happening.

## What are you changing?

Re: Celery Performance tuning


sastels (approved):
LGTM!

PR #1941, opened 2023-08-08
Setting staging HPA tuning to 12 pods per cycle at 45 seconds

ben851:
Set celery HPA tuning to 12 pods per iteration at 45 seconds per. 

## What are you changing?

Re: Celery perf tuning


sastels (approved):
don't like using 45 seconds but ok fine.

PR #1954, opened 2023-08-17
Moving Service Annotations to Kustomize Patch

ben851:
Updating the annotation configuration for internal load balancers to be part of environment patches, so we can continue to release to production without completing this migration

## What are you changing?

Re: EKS Load Balancers


sastels (approved):
LGTM!

PR #1966, opened 2023-08-24
chore: send a deployment log to sentinel

CalvinRodo:

This is being as part of the tech authority work that core services is partaking in and is also a metric that Platform currently must report on. 

This uses CDS' sentinel-forward-data-action to send a notification to sentinel that we have deployed notify.

This will be used for tracking the Deployment Frequency and is related to the following issues.

- https://github.com/cds-snc/platform-core-services/issues/437
- https://github.com/cds-snc/platform-core-services/issues/352


jimleroyer (approved):
LGTM

PR #2011, opened 2023-09-19
feat: Upgrading cwagent in prod

ben851:

CWAgent and CWAgent for prometheus will be updated to the latest version in production.

## What are you changing?

In addition to keeping up to date with patches, this is also in attempt to address periodic metric dropping in cloudwatch.

## After merging this PR


jimleroyer (approved):
Let's check the performance tests numbers the day after this merge. Curious if that can have any impact but also, we need to adopt this habit to see if we have side effects.

PR #2022, opened 2023-09-20
Adding retry logic to kustomize apply

ben851:
The merge to staging github action will now use a shell script that contains retry logic when doing kubectl apply -k. This is required for karpenter because kustomize does not have a mechanism to wait for CRD creation or services entering ready state before proceeding to the next resource. 

This script will attempt to apply Kustomize. If it works, it will break out of a loop. If it fails, it will wait 10 seconds and retry. It will retry up to 5 times before erroring out fully.

## What are you changing?

In support of K8s performance tuning and karpenter implementation


scripts/applyKube.sh
```diff
@@ -0,0 +1,11 @@
+retryCount=0
+workDir=$1
+kubeConfig=$2
+pushd $workDir
```
sastels, line 4: nice

sastels (approved):
LGTM

scripts/applyKube.sh
```diff
@@ -0,0 +1,11 @@
...
+until [ "$retryCount" -ge 5 ]
+do
+   kubectl apply -k . $kubeConfig  && break
```
jimleroyer, line 7: Interesting way to break out of the loop, my shell programming is way far behind but that looks quite idiomatic to the language.

jimleroyer (approved):
Approved with a wish for some documentation on the new script.

PR #2043, opened 2023-09-26
Reorganizing kustomize folders

ben851:

This is a code structure change and should not affect the code that is deployed to production or staging.

By organizing our code in sub-folders, we will be able to control what features are deployed to each environment. This is useful for example, for reducing costs in scratch accounts, since we can choose to disable cloudwatch agents for that environment only.

## What are you changing?
Code organizational structure

Part of efforts to reduce costs in dev environments.


env/scratch/kustomization.yaml
```diff
@@ -1,20 +1,26 @@
 bases:
-  - ../../base
-  - ../../base-karpenter
-
+  - ../../base/karpenter
+  - ../../base/kube-state-metrics
+  #- ../../base/prometheus-cloudwatch
```
sastels, line 4: Do we still want these commented out or should we straight up delete the lines? I guess it's useful to have the commented out lines for when we want to include and tests this code in scratch?

sastels (approved):
good idea! one question around some commented out lines.

Suggest not merging until tomorrow after release so if prod breaks we know whose fault it is 😅

env/scratch/kustomization.yaml
ben851: Yeah I have them commented just so I can easily add them back when I need to test those specific functions

PR #2046, opened 2023-09-27
removing memory bounds for celery hpa in prod

ben851:
Celery HPA in prod will no longer scale based on memory. 

## What are you changing?

Re: Celery performance tuning


sastels (approved):
Thanks for the catch!

PR #2060, opened 2023-10-03
Adding k8s event logger retention

ben851:
A new deployment, k8s event logger, will be created to ingest k8s events into cloudwatch.

## What are you changing?

Re: Debugging api replicas unavailable.


sastels (approved):
I like how easy it is to just add modules like this to staging/scratch 👍

PR #2062, opened 2023-10-04
adding pod disruption budget for api

ben851:
A pod disruption budget for the api pods in kubernetes will be set. We will not allow more than two pods to be down at once. This will hopefully force karpenter to be patient when shuffling things around.

## What are you changing?

Re: Performance tuning


sastels (approved):
:shipit:

PR #2067, opened 2023-10-04
setting prod celeries to use karpenter nodes

ben851:
Both email and sms send celery will use karpenter nodes.

## What are you changing?

Performance tuning

### DIFF
--- a/env/production/node-selector-patch.yaml
+++ b/env/production/node-selector-patch.yaml
@@ -2,10 +2,6 @@
 
 #### KARPENTER SPOT INSTANCES - EPHEMERAL, STATE NOT REQUIRED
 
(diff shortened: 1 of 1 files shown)


sastels:
It looks like things got reordered a bit, but in the end we have everything ON_DEMAND except celery / celery-sms-send which are now on spot?

ben851:
> It looks like things got reordered a bit, but in the end we have everything ON_DEMAND except celery / celery-sms-send which are now on spot?

Yeah, I copy/pasted from staging which I guess had a slightly different order, I figure good to keep them all in the same order for the future.

You are correct that only the two celery sends will go to spot!

sastels (approved):
🚀

PR #2079, opened 2023-10-11
Adding the k8s event logger to production

ben851:
The k8s event logger will be deployed to production so that we can keep a history of the events that go on in kube system.

## What are you changing?

Re: Performance tuning but also general reliability of Notify.


jimleroyer (approved):
I'd like to see that in a code review session. 👍

PR #2094, opened 2023-10-16
Setting production to run celery on karpenter nodes

ben851:
When released, celery will use Karpenter nodes again.

We have introduced a cwagent health check script that should fix the intermittent metric send errors.

## What are you changing?

Re: Celery performance tuning and node autoscaling


jimleroyer (approved):
LGTM

PR #2095, opened 2023-10-16
Moving API off of karpenter nodes for now while we look into 502s

ben851:
In staging, the API will now deploy to standard k8s nodes instead of karpenter spot instances. This is to get rid of the 502 errors we're intermittently seeing. A new card will be opened to debug that.

## What are you changing?


jimleroyer (approved):
LGTM

PR #2098, opened 2023-10-18
Setting system logging components to be node cricial priority

ben851:
We will set the logging system components to be node critical priority (IE they have the authority to kick off other pods in the event that there aren't enough resources). 

## What are you changing?

Re: Node autoscaling and performance tuning.


jimleroyer (approved):
LGTM

PR #2113, opened 2023-10-25
[MANIFEST] Setting HPA for celery SMS to 25% CPU. Adjusting pod count

ben851:
Setting HPA for Celery SMS in prod to 25% CPU Threshold and re-enabling autoscaling

## What are you changing?


sastels (approved):
Tested in staging 👍

PR #2119, opened 2023-10-31
R5 only

ben851:
Celery Karpenter will only use r5.large and r5.xlarge in staging and prod. No more m5's as they don't seem to behave.

## What are you changing?

Node autoscaling


andrewleith (approved):
LGTM

PR #2121, opened 2023-10-31
Re-enabling consolodation

ben851:

Staging will have node consolidation re-enabled to save on the $$

Part of node autoscaling work


sastels (approved):
🎉

PR #2125, opened 2023-11-01
fix: Setting higher memory limits in cwagent

ben851:

The memory limit for cwagent in kubernetes will be set to 1gb from 200mb.

## What are you changing?

We are experiencing OOM crashes on cwagent.

## After merging this PR


sastels (approved):
👍 assuming we have the memory available?

PR #2128, opened 2023-11-02
Karpenter spot instances on imdsv1, celery back on spot

ben851:

Karpenter instances will now you imdsv1 instead of v2 to conform with the on demand nodes and not cause fluentbit errors.

## What are you changing?

Re: incident on celery metric timeout errors


sastels (approved):
LGTM 🤞

PR #2148, opened 2023-11-08
Staging fluentbit membufflimit 1

ben851:

Tripling membuff limit on staging celery fluentbit in order to alleviate emitter paused errors under load.
Increasing maximum memory usage for the daemonset to accommodate.

## What are you changing?

Errors in fluentbit when under load


jimleroyer (approved):
Let's try this out.

PR #2149, opened 2023-11-08
Fluenbit Pipeline Aliases and Celery MLP emitter Mem Buff Limit

ben851:
Celery in staging will have aliases assigned to the pipeline stages to better troubleshoot issues.

Celery MLP emitter will have an increased membuff limit to avoid lost chunks under load.

### DIFF
--- a/env/staging/fluentbit.yaml
+++ b/env/staging/fluentbit.yaml
@@ -81,6 +81,7 @@ data:
   notify-log.conf: |
     [INPUT]
         Name                tail
+        Alias               notify-app-input-tail 
         Tag                 application.*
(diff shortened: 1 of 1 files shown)


env/staging/fluentbit.yaml
```diff
@@ -155,28 +161,32 @@ data:
...
+        Labels                Off
+        Annotations           Off
+        Use_Kubelet           On
+        Kubelet_Port          10250
+        Buffer_Size           0
+        emitter_mem_buf_limit 150MB
 
     [FILTER]
         name                  multiline
+        Alias                 notify-celery-multiline-filter
```
jimleroyer, line 182: I am 😨  seeing this change.

jimleroyer (approved):

env/staging/fluentbit.yaml
ben851: Bad memories!

PR #2170, opened 2023-11-09
Putting membuff limit back

ben851:
Putting the upgraded memory buffer limits again!


patheard (approved):
LGTM!

PR #2194, opened 2023-11-17
[MANIFEST] Adding disruption budgets for celery

ben851:
We will set pod disruption budgets for celery to try and make sure that we don't bring down all pods at once

## What are you changing?

Re: Incident - sms pods going to 0


sastels:
Note that this will also release #2179

sastels (approved):

PR #2199, opened 2023-11-17
Moving cwagent check outside of celery

ben851:
The cloudwatch agent init check will be moved to an init container so that kubernetes is aware of the true status of the celery pod.

## What are you changing?

Re: Incident 2023-11-17-pod-down-sms-delay

### DIFF
--- a/base/notify-celery-email-send/celery-email-send-deployment.yaml
+++ b/base/notify-celery-email-send/celery-email-send-deployment.yaml
@@ -21,6 +21,37 @@ spec:
       labels:
         app: celery-email-send
     spec:
+      initContainers:
+      - name: wait-cwagent-ready
+        env:
+          - name: STATSD_HOST
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+        image: api
+        imagePullPolicy: IfNotPresent
+        command: ["sh", "-c"]
+        args:
+          - >-
+            if [ -n "${STATSD_HOST}" ]; then
+                echo "Initializing... Waiting for CWAgent to become ready within the next 30 seconds."
+                timeout=30
+                while [ $timeout -gt 0 ]; do
+                    if nc -vz "$STATSD_HOST" 25888; then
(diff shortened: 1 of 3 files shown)


base/notify-celery-email-send/celery-email-send-deployment.yaml
```diff
@@ -21,6 +21,37 @@ spec:
...
+      initContainers:
+      - name: wait-cwagent-ready
+        env:
+          - name: STATSD_HOST
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+        image: api
+        imagePullPolicy: IfNotPresent
+        command: ["sh", "-c"]
```
jimleroyer, line 33: Any advantage over having the scripts inlined rather than executing a file? Because that was the quick solution as the existing scripts are in the API repo?

ben851: Last week it didn't occur to me to just use the api image, and I didn't want to build a new image for a single thing...

We can move this to scripts, it will just require two PRs instead of one. You're probably right in terms of keeping things consistent.

base/notify-celery-email-send/celery-email-send-deployment.yaml
```diff
@@ -21,6 +21,17 @@ spec:
...
+      - name: wait-cwagent-ready
+        env:
+          - name: STATSD_HOST
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+        image: api
+        imagePullPolicy: IfNotPresent
+        command: ["/bin/sh"]
+        args: ["-c", "sh /app/scripts/cwcheck.sh"]
```
jimleroyer, line 34: 🥳

jimleroyer (approved):
That looks good. One more request unrelated to the code itself: please let's explain the root cause of our change to this PR's description because future code archeologist wondering why we have this init container will be fast tracked on it.

PR #2200, opened 2023-11-17
[MANIFEST] Added env var targeting newrelic config for the admin

jimleroyer:
Reverts cds-snc/notification-manifests#2195

This is putting back the NewRelic TF changes back into the mainline which was previously merged in #2179.

Let's wait to make a manifests components release before merging this one again.


sastels (approved):
LGTM but as you say let's hold off on merging this until we do an admin release so we don't block the release pipeline

whabanks (approved):

PR #2206, opened 2023-11-20
[MANIFEST] Turning off TLS verification on kubernetes filter

ben851:
The Fluent Bit kubernetes filter will have TLS verification disabled when talking to the kubelet. TLS is still enforced, but we do not verify the certificate... This is to test the TLS verification error we are receiving on fluent bit startup.

## What are you changing?

Re: TLS errors in fluent bit.

### DIFF
--- a/env/production/fluentbit.yaml
+++ b/env/production/fluentbit.yaml
@@ -331,10 +331,7 @@ spec:
         version: v1
         kubernetes.io/cluster-service: "true"
     spec:
-      initContainers:
-      - name: wait-for-init
-        image: busybox:1.28
-        command: ['sh', '-c', 'echo "Waiting for 2 seconds for node to sort itself out" && sleep 2']      
+      priorityClassName: system-node-critical
       containers:
       - name: fluent-bit
         image: public.ecr.aws/aws-observability/aws-for-fluent-bit:stable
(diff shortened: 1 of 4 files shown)


env/production/fluentbit.yaml
```diff
@@ -331,10 +331,7 @@ spec:
         version: v1
         kubernetes.io/cluster-service: "true"
     spec:
-      initContainers:
-      - name: wait-for-init
-        image: busybox:1.28
-        command: ['sh', '-c', 'echo "Waiting for 2 seconds for node to sort itself out" && sleep 2']
```
jimleroyer, line 337: oh wait we get rid of the 2 seconds waiting here? so we can see the tls error and investigate?

sastels: and do we want this in prod too or just staging?

ben851: The interwebs and SO say that this is pretty much the solution to getting rid of this error... I want to see if it does in our case. 

As for removing the wait in production, 2 seconds wasn't long enough so I got rid of it... I can add it back or even up it to 10 seconds here and see if that helps for prod

jimleroyer: Let's test this in staging and see how the experiments goes. We can revert back to what worked best then. 👍

jimleroyer (approved):

PR #2209, opened 2023-11-21
Setting min available for celery pdb to 2

ben851:
Celery pod disruption budget will be set to two, so that there will always be at least one allowed disruption when evicting pods.

## What are you changing?

Re: Action item from failed TF worker upgrade.


jimleroyer (approved):
LGTM

PR #2216, opened 2023-11-23
Node Affinity Dev/Staging

ben851:

Attempting to split the distribution of celery pods across spot and on demand instances.

## What are you changing?

Node and Celery auto scaling


jimleroyer (approved):
Let's test this 🥼

PR #2225, opened 2023-11-28
Pod skew dev

ben851:
Celeries will be split between primary and karpenter nodes in production

## What are you changing?

Re: Node autoscaling


env/production/node-selector-patch.yaml
```diff
@@ -13,8 +13,23 @@ metadata:
...
+            preference:
+              matchExpressions:
+              - key: eks.amazonaws.com/capacityType
+                operator: In
+                values:
+                - ON_DEMAND
+          - weight: 50
+            preference:
+              matchExpressions:
+              - key: karpenter.sh/capacity-type
```
jimleroyer, line 29: Ok I didn't get that: we are letting amazon eks manage the on demand, and karpenter managing the spot instances. Is that right?

ben851: Well the on-demand instances are really managed by us... as in there's still no autoscaling of the primary node group... just the karpenter ones... But this PR will send some celery pods to the on demand instances so that we can be sure that some celery will always be running

jimleroyer: Yes that makes sense. I meant eks manage this, we're doing it the more traditional way of Kubernetes (but with no hpa for this). We can sync if you think my understanding is wrong.

ben851: The HPA still exists but K8s will look at the call for more pods by the HPA and (try to) distribute them between spot instances and primary nodes according to these weights

PR #2234, opened 2023-11-29
DD API Node Affinity Fix - Staging

ben851:
Missed the last line from copy/paste


sastels (approved):
opppppss

PR #2241, opened 2023-12-04
Notify PriorityClasses and Pod Disruption Budgets

ben851:

When merging, we will create priority classes that allow pods that are intended to run on primary nodes only (admin/celery-beat/doc-download etc) to evict any celery pods that are on the primary nodes if they need to claim resources.

I also added pod disruption budgets to all notify items to ensure we always have at least one of everything running.

I also corrected the naming on the existing pod disruption budgets which were called hpa's for some reason (in name only)

FINALLY - I set K8s api to prefer primary nodes but also allowed it to run on spot instances, since we don't really need this up anyway.

## What are you changing?

Part of node autoscaling


base/notify-celery-main/celery-deployment.yaml
```diff
@@ -22,7 +22,8 @@ spec:
       labels:
         app: celery
         # profile: fargate
-    spec:
+    spec:  
+      priorityClassName: celery
```
jimleroyer, line 26: Does that mean that some celery workers could be in the same situation of 0 instances if nodes are reworked during a scaling procedure?

ben851: In theory yes, but I think the risk is low because we only ever have 2 instances of things like admin or doc download running, we won't run into a scenario where we need to evict every celery pod from primaries (right now each primary node is running ~4 celery pods)

I'd like to give it a try this way, with Steve's tests. If it proves to be too unpredictable, I'll do the two deployment thing.

sastels: sounds good.

sastels (approved):
Let's give it a test!

PR #2243, opened 2023-12-05
Modifying scaling behaviour for celeries

ben851:
Changing the scale up chunks for the celeries to scale up slower, and not overload the system.

Also adjusted celery email CPU usage target back to 50 because we are seeing 30 pods even when there is low load on the system.

To be tested in staging.

## What are you changing?

Celery and node autoscaling

### DIFF
--- a/env/production/performance.yaml
+++ b/env/production/performance.yaml
@@ -113,6 +113,14 @@ spec:
         averageUtilization: 50
         type: Utilization
     type: Resource
+  behavior:
+    scaleUp:
+      stabilizationWindowSeconds: 0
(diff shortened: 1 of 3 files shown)


env/production/performance.yaml
```diff
@@ -159,14 +167,14 @@ spec:
   - resource:
       name: cpu
       target:
-        averageUtilization: 25
```
sastels, line 162: Are we confident that the system will still scale **up** appropriately? Do we want to test 50 on staging first?

sastels (approved):
woot!

PR #2247, opened 2023-12-06
[MANIFEST] Added sensitive PT services

jzbahrai:
tags manifest to a new patch release and deploys it .

this PR adds few sensitive services.

when I run this: ```make diff-production```
I see the change, but I dont want to post service ids here


andrewleith (approved):
LGTM - verified the 2 service ids being added.

PR #2265, opened 2023-12-12
fix: staging celery core request 300m -> 150m

sastels:

## What are you changing?

Celery pods are not scaling up fast enough to process callbacks (and not even getting to max pods)


ben851 (approved):
LGTM

PR #2273, opened 2023-12-18
Removing old scratch context

ben851:
An old scratch context script was accidentally included, which had a hard-coded (but not active) password for grafana in my scratch account. Removing.


sastels (approved):
So this should have no affect on staging / prod, correct?

ben851:
> So this should have no affect on staging / prod, correct?

Nope, not used

PR #2280, opened 2024-01-04
Adjusting resource request/limits

ben851:

Adjusting resource request/limits in order to not have a bunch of wasted compute and memory. Staging only at this point so that I can run some load tests.

## What are you changing?

Performance tuning/cost saving


andrewleith (approved):
lgtm!

PR #2290, opened 2024-01-09
fix: Tweak celery scaling on dev and staging

sastels:

## What are you changing?

celery-scalable not scaling up even though there are 40k tasks in the delivery-receipts queue. This should help.


ben851 (approved):
LGTM

PR #2291, opened 2024-01-09
Reconfiguring dev to make it actually work

ben851:
Updating the dev kustomize to work with the new cluster.


sastels (approved):
Let's test it out!

PR #2293, opened 2024-01-10
fix: Add tasks to update lambda based on images

jzbahrai:
Added changes to update the lambda images for system_status and heartbeat

### DIFF
--- a/.github/workflows/merge_to_main_production.yaml
+++ b/.github/workflows/merge_to_main_production.yaml
@@ -10,6 +10,8 @@ on:
 env:
   PRIVATE_ECR: ${{ secrets.PRODUCTION_AWS_ACCOUNT }}.dkr.ecr.ca-central-1.amazonaws.com/notify
   API_LAMBDA_IMAGE: api-lambda:a4e4b54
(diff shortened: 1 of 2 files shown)


.github/workflows/merge_to_main_production.yaml
```diff
@@ -10,6 +10,8 @@ on:
 env:
   PRIVATE_ECR: ${{ secrets.PRODUCTION_AWS_ACCOUNT }}.dkr.ecr.ca-central-1.amazonaws.com/notify
   API_LAMBDA_IMAGE: api-lambda:a4e4b54
+  HEARTBEAT_LAMBDA_IMAGE: heartbeat:fb6dcdfda3c8e21ef997e75d1cb872bb08916cf3
+  SYSTEM_STATUS_LAMBDA_IMAGE: system_status:fb6dcdfda3c8e21ef997e75d1cb872bb08916cf3
```
jimleroyer, line 14: We're setting these manually on first iteration and automatically later?

jimleroyer (approved):

.github/workflows/merge_to_main_production.yaml
jzbahrai: yes

PR #2340, opened 2024-01-25
Nginx, helmfile, hasura ingress for VPN

ben851:

**[Depends on this notification-terraform PR](https://github.com/cds-snc/notification-terraform/pull/1133)**

This is a big one:
- Adding nginx ingress controller to k8s to control the ingress routes of our utilities (ex hasura and maybe blazer eventually?)
- Adding the K8s CSI driver, and the associated AWS provider to K8s so that we can load secrets directly into k8s from AWS SSM
- The above is done using helm/helmfile because it is much less work, and can provide a safe test of how helmfile works
- Because of helmfile, adding a new job on PR create that does a helmfile diff.
- Because of helmfile, adding a new job on merge to main that does helmfile apply

## What are you changing?

Ben going crazy


helmfile/getContext.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
jimleroyer, line 1: Let's add a file header to explain the usage of this file. 🙏

helmfile/helmfile.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - ./overrides/system/nginx.yaml.gotmpl
+
+  - name: ingress-secrets
+    namespace: nginx
+    disableValidationOnInstall: true
+    labels:
+      app: nginx
+      tier: secrets
+      category: system
+    chart: adfinis/csi-secret-provider-class
```
jimleroyer, line 42: Let's have a `version` field for all referenced charts, for security purpose. 🙏

.github/workflows/helmfile_staging_apply.yaml
```diff
@@ -0,0 +1,42 @@
...
+          # Fetches entire history, so we can analyze commits since last tag
+          fetch-depth: 0
+      - name: Configure kubeconfig
+        run: |
+          aws eks update-kubeconfig --name notification-canada-ca-staging-eks-cluster        
+      - name: Load EnvVars
+        run: |
+            ./helmfile/getContext.sh true
+      - name: Run helmfile
+        id: helmfile_diff
```
jimleroyer, line 39:
```suggestion
        id: helmfile_apply

File: helmfile/helmfile.yaml (hunk 2 above)
Comment (ben851) [reply helmfile/helmfile.yaml]: This actually gets removed with latest changes since the chart is now local 🥳

File: helmfile/getContext.sh (hunk 1 above)
Comment (ben851) [reply helmfile/getContext.sh:1]: Done!

File: .github/workflows/helmfile_staging_apply.yaml [hunk 4]
```diff
@@ -0,0 +1,44 @@
...
+  STAGING_AWS_ACCOUNT: ${{ secrets.STAGING_AWS_ACCOUNT }}
+
+jobs:
+  helmfile-diff:
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Inject token authentication
+        run: |
+          git config --global url."https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/".insteadOf "https://github.com/"
```
patheard, line 20: Minor comment here, if you wanted to move away from a hard-coded token, you could replace this with an SRE GitHub app's temp token:
https://github.com/cds-snc/gcds-components/blob/424b138d6ebf8e3df77b2e265ff386bf1e39d7cf/.github/workflows/release-generator.yml#L15-L19

ben851: Nice, not opposed to this but I think it's out of scope for this particular PR - I'll create a card to migrate all workflows to this method.

ben851: Card created:
https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/292

helmfile/helmfile.yaml
```diff
@@ -0,0 +1,51 @@
...
+releases:
+
+  - name: ingress
+    namespace: nginx
+    disableValidationOnInstall: true
+    labels:
+      app: nginx
+      tier: main
+      category: system
+    chart: charts/nginx-ingress
```
patheard, line 26: Just curious, but why a local chart in this case?

jimleroyer (approved):
LGTM, it's only dev and staging env at the moment.

Things to follow up with:

* hard-coded token
* overriding versions contained in the helm charts and check if these could get managed by renovate

PR #2342, opened 2024-01-26
Prevent naming variations of .env files to be submited to git

jimleroyer:
# Summary | Résumé

Prevent naming variations of .env files to be submited to git.

# Test instructions | Instructions pour tester la modification

Create a file named `.env.staging` locally and check if it can get submitted with git.


whabanks (approved):
LGTM, tested that modifications to `dev|prod|staging/.env.zip.enc.aws` still take.

sastels (approved):
tested locally, lgtm!

PR #2350, opened 2024-01-31
fix: Remove FF_EMAIL_DAILY_LIMIT feature flag

whabanks:
# Summary | Résumé

Remove `FF_EMAIL_DAILY_LIMIT` feature flag. 

# Test instructions | Instructions pour tester la modification

Decrypt .env configs and verify that only `FF_EMAIL_DAILY_LIMIT` was changed for each environment.

### DIFF
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -25,7 +25,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@75cb7af1033cfb77c9fc7d8abc30420008f558f4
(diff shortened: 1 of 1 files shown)


jzbahrai:
Ran this on staging and prd and got the following:

```
Archive:  .previous.env.zip
  inflating: .env
Archive:  .env.zip
  inflating: .env
47d46
< FF_EMAIL_DAILY_LIMIT=true
```

whabanks:
> thanks for running the linter! I would suggest doing that in another PR next time, just so we can have separations of concerns (especially as it was many changes, if it was one of two that would be okay imo)

Good point, I have vscode set up to run the linter on any file that I ctrl + S. I'll disable that going forward and keep things separate.

PR #2369, opened 2024-02-09
Blazer almost fully working in dev

ben851:
We will deploy a blazer instance into Kubernetes, and it will be accessible by:

https://blazer.staging.notification.internal/

This does not remove the existing blazer ECS!

## What are you changing?

VPN Tools

### DIFF
--- a/helmfile/overrides/tools/blazer.yaml.gotmpl
+++ b/helmfile/overrides/tools/blazer.yaml.gotmpl
@@ -1,0 +1,136 @@
+applicationName: blazer
+
+ingress:
+  enabled: true
+  ingressClassName: nginx
+  servicePort: http
+  hosts:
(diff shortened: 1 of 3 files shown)


helmfile/overrides/tools/blazer.yaml.gotmpl
```diff
@@ -0,0 +1,136 @@
+applicationName: blazer
+
+ingress:
```
jimleroyer, line 3: To have the reverse proxy setup via nginx, we need to define the ingress, and no port forwarding via a service, if I got this correctly?

jimleroyer (approved):
Looks okay to me

helmfile/overrides/tools/blazer.yaml.gotmpl
ben851: You got it!

PR #2383, opened 2024-02-20
fix: update sensitive services list

andrewleith:
Add a new service to the list, as per [conversation here](https://gcdigital.slack.com/archives/CV38DBNVA/p1707923663881079).


sastels (approved):
checked the file locally :shipit:

PR #2417, opened 2024-03-06
Deploy workflow tuning

ben851:
K8s Rollout Workflow take 2

I'm testing the remote launch and passing the docker_sha as input, hence why I've commented everything else out - I just want to see that the value gets accepted


sastels (approved):
Let's test it out!

PR #2433, opened 2024-03-07
Stand Alone API K8s Rollout Workflow

ben851:
## What Happens

A new github actions workflow will be created that can be remotely triggered by other github actions. This action will update the API/Celery image definition in K8s and do a rollout/restart of the app.

## After merging this PR


sastels (approved):
🚀

PR #2438, opened 2024-03-07
Set the docker tag in shell for k8s workflow

ben851:
The K8s API Rollout Workflow will get the docker tag for the API from the input.

## What are you changing?
Github config

Re: Private EKS config

- [K] I know how to get kubectl credentials in case it catches on fire

## After merging this PR


sastels (approved):
sure

PR #2448, opened 2024-03-12
chore: Fixing helmfile staging and reverting merge to main staging

ben851:
I somehow deleted the helmfile install step on staging helmfile apply. Put that back in.

Also, the mergeToMainStaging workflow is using the sentinel forwarder which I missed when doing the internal ARC. I'm reverting that one for now.

## What are you changing?

Re: Private EKS cluster

## After merging this PR


sastels (approved):
let us test

PR #2452, opened 2024-03-12
chore: Fix kubectl in merge to main staging

ben851:
Merge to main staging was failing on installing kubectl when run privately. I've swapped the install method to use the same method that helmfile uses to install kubectl. This should resolve the issue.


sastels (approved):
could add your "installing k8s" comment here if you wish

PR #2459, opened 2024-03-13
chore: Kustomize update

ben851:

The kustomize code has been refactored to align with the latest version best practices. This involves separating all of the patches out into single files. 

The production and staging workflows have been updated to latest kubectl

The staging workflow has also been set to run internally

I have also removed the scratch folder, we can re-use dev for this.

## What are you changing?

Re: private EKS cluster

- [X] I made sure that the changes are as expected in [Notify staging](https://staging.notification.cdssandbox.xyz/)

## After merging this PR


jimleroyer (approved):
We spoke offline and decided to contain the changes to non-production environments for now, and let it simmer in staging. There should be no changes value-wise, but just in case that we missed something. ✔️

PR #2460, opened 2024-03-13
chore: Update catalog-info file

gcharest:

Update the catalog info file to:
- assign the notify-dev team as owner of the component

This is to help :
- automate the discovery of all the CDS software components and systems
- register components to benefit from various enhancement capabilities the dev portal will offer over time.


catalog-info.yaml
```diff
@@ -10,4 +10,15 @@ metadata:
 spec:
   type: website
   lifecycle: experimental
```
jimleroyer, line 12:
```suggestion
  lifecycle: production
```

catalog-info.yaml
```diff
@@ -10,4 +10,15 @@ metadata:
 spec:
   type: website
```
jimleroyer, line 11: What are the possible types? 🤔

jimleroyer:
Wondering if we can put a header comment to explain the catalog file refers to, the goal and how to access it?

```
---
# Metadata for the backstage catalog accessible at this link: ...
---
```

or something along these lines.

catalog-info.yaml
jimleroyer: This would be for the lifecycle though, right? For type, so far, I saw `website`, `service` and `library`. Would there be other ones?

gcharest: Apologies, that's right, although you can also set set any string if that's needed by the org. so if we want to set a custom one for our own use, that's totally acceptable.

The default ones generate specific views in the catalog component so we can tailor the way we display each type:

https://github.com/cds-snc/backstage/blob/9c80931545ceb00774e3006a896f3d54d9ee1e4c/packages/app/src/components/catalog/EntityPage.tsx#L415

jimleroyer (approved):
LGTM!

PR #2486, opened 2024-03-21
Forcing fluent bit to be system critical

ben851:
Fluent bit will be system critical and will evict pods to make sure it's running.

## What are you changing?

While perf testing, we noticed that fluentbit didn't always find a spot on its node because celery took too many resources. 

## After merging this PR


P0NDER0SA (approved):
looks about right.

PR #2506, opened 2024-03-27
feat(branding): turn  FF_NEW_BRANDING ON

andrewleith:

This PR turns on the branding feature flag (`FF_NEW_BRANDING = true`)

```
AWS_PROFILE=notify-prod make diff-production   
Archive:  .previous.env.zip
  inflating: .env                    
Archive:  .env.zip
  inflating: .env                    
47c47
< FF_NEW_BRANDING=false
---
> FF_NEW_BRANDING=true
```


whabanks (approved):
LGTM, change is as expected with diff'd

PR #2510, opened 2024-04-02
Production k8s rollout workflows

ben851:

The workflows for updating notify images in k8s in production will be added. They will not be "active" - that requires separate PRs in the corresponding repository (API, DD-API, Documentation, Admin). 

Once this is merged, I will try manually running them to verify that they work (I will have them redeploy the current version)  before making changes to the other repositories.

## What are you changing?
Github Workflows only

Preparation for private Prod EKS cluster.

## After merging this PR


sastels (approved):
ok I think this is good to put in and test!

PR #2538, opened 2024-04-17
Adding workflow dispatch to helmfile runs

ben851:
Adding workflow dispatch to helmfile runs so that we can manually run them

## What are you changing?
- [ ] Github workflows 

Release not working


andrewleith (approved):
LGTM

PR #2540, opened 2024-04-17
Helmfile single apply tag

ben851:
Setting the individual apply to set the docker_tag env variable. Setting the corresponding helmfile override to also use docker_tag

## What are you changing?

Re: github arc docker build

## After merging this PR


P0NDER0SA (approved):
LGTM

PR #2627, opened 2024-05-22
feat: add AWS_PINPOINT_DEFAULT_POOL_ID variable to k8s and set in staging

sastels:
## What happens when your PR merges?
    - `feat:` - tag `main` as a new minor release

## What are you changing?
- [ ] Releasing a new version of Notify
- [x] Changing kubernetes configuration

## Provide some background on the changes

Following https://github.com/cds-snc/notification-api/pull/2173, we add the new `AWS_PINPOINT_DEFAULT_POOL_ID` variable to k8s.

```
$ make diff-staging
> AWS_PINPOINT_DEFAULT_POOL_ID=pool-8885654b47d6466f9aaee14c62494ff2
```

```
$ make diff-production
> AWS_PINPOINT_DEFAULT_POOL_ID=
```

Note that in production (where `AWS_PINPOINT_SC_POOL_ID` and `AWS_PINPOINT_DEFAULT_POOL_ID` are not set) we will still fall back to SNS.

## Checklist if making changes to Kubernetes:
- [x] I know how to get kubectl credentials in case it catches on fire

PR #2634, opened 2024-05-22
Revert "feat: add AWS_PINPOINT_DEFAULT_POOL_ID variable to k8s and set in staging"

sastels:
Reverts cds-snc/notification-manifests#2627

turning on pinpoint broke the fake callbacks generated when sending to the fake phone number 😞 

I'll fix that, then bring this PR back.


jimleroyer (approved):
LGTM

PR #2636, opened 2024-05-24
Revert "Revert "feat: add AWS_PINPOINT_DEFAULT_POOL_ID variable to k8s and set in staging""

sastels:
Reverts cds-snc/notification-manifests#2634

Use pinpoint by default.

After merging #2627 we found that the fake internal test number caused an error. We fix this in [Fix fake callback](https://github.com/cds-snc/notification-api/pull/2178), so after that PR merges we can turn Pinpoint back on.

```
make diff-production
> AWS_PINPOINT_DEFAULT_POOL_ID=
```

```
make diff-staging
> AWS_PINPOINT_DEFAULT_POOL_ID=pool-8885654b47d6466f9aaee14c62494ff2
```


jimleroyer (approved):
Revert of revert should be good once https://github.com/cds-snc/notification-api/pull/2178/files is merged.

PR #2644, opened 2024-05-29
Update list of sensitive services

andrewleith:
# Summary

Add new services to the sensitive services list.


jzbahrai (approved):
Ran:
```
make diff-production
Archive:  .previous.env.zip
  inflating: .env
Archive:  .env.zip
  inflating: .env
  ```
  
  not posting all the IDs but there was a diff

PR #2655, opened 2024-06-04
Karpenter AMI set via secret

ben851:

The AMI ID for karpenter nodes will now be set via a secret set in terraform so that we can manage the AMI updates in one place. 

## What are you changing?

This was discovered while upgrading to K8s 1.30 - the karpenter spot nodes were still using the old 1.29 AMI.

## After merging this PR


sastels (approved):
nice

PR #2668, opened 2024-06-06
feat: add AWS_PINPOINT_DEFAULT_POOL_ID to k8s

sastels:

## What are you changing?

Add the new `AWS_PINPOINT_DEFAULT_POOL_ID` env var to k8s (celery sms sending pods). Initially set the value to empty on both staging and production.
```
$ make diff-staging
> AWS_PINPOINT_DEFAULT_POOL_ID=
```

```
$ make diff-production
> AWS_PINPOINT_DEFAULT_POOL_ID=
```

We had a few issues making these changes before but I believe the problems have been sorted out. Local testing with
```
AWS_PINPOINT_DEFAULT_POOL_ID=
```
works, and this PR uses an empty `AWS_PINPOINT_DEFAULT_POOL_ID` for both staging and production, so we should be able to verify this configuration on staging before deploying it to production.

## After merging this PR


env/staging/.env.zip.enc.aws
jimleroyer, line 1: So ideally, this is where this information would rather be contained in the database, leveraging what the app team will do, right?

jimleroyer:
@sastels The encrypted values are empty at the moment for both staging and production, from this PR?

env/staging/.env.zip.enc.aws
sastels: Right now I just added a template list to these environment variables, but in the future we could decide more dynamically what to send with the shortcode (ie going by template priority, for example)

jimleroyer: I just saw an empty variable when I checked for both of these locally. Is that normal?

sastels: The `AWS_PINPOINT_DEFAULT_POOL_ID` is empty on staging and prod, but the `AWS_PINPOINT_SC_POOL_ID` and `AWS_PINPOINT_SC_TEMPLATE_IDS` variables have values on both. So currently we use Pinpoint for the `AWS_PINPOINT_SC_TEMPLATE_IDS` template but SNS for the rest. This won't change with this PR, but next we'll set the `AWS_PINPOINT_DEFAULT_POOL_ID` variable and use the Pinpoint default pool for the rest.

jimleroyer: I looked too quickly and did not see these, you are right and these have values!

jimleroyer (approved):

PR #2710, opened 2024-06-26
Github Actions using VPN In Staging

ben851:
All github actions that require connectivity to EKS in staging will now use openvpn to connect to the cluster instead of running on self-hosted github actions.

## What are you changing?

Working on stabilizing and simplifying our workflows.

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document download API

## After merging this PR


.github/workflows/config/staging.ovpn
ben851, line 1: Not happy that this has to live in code, but it looks like it's the only way at the moment. This does not contain the private key, so it should be secure.

.github/workflows/admin-rollout-k8s-staging.yaml
```diff
@@ -44,6 +44,17 @@ jobs:
...
+      - name: Install OpenVPN
+        run: |
+          sudo apt update
+          sudo apt install -y openvpn openvpn-systemd-resolved
+
+      - name: Connect to VPN
+        uses: "kota65535/github-openvpn-connect-action@v2"
+        with:
+          config_file: .github/workflows/config/staging.ovpn
+          client_key: ${{ secrets.STAGING_OVPN_CLIENT_KEY }}
```
jimleroyer, line 56: Could we have this flow from our 1pw stuff? just wondering here, no blocker

jimleroyer (approved):
LGTM, left a question but no blocker

.github/workflows/admin-rollout-k8s-staging.yaml
ben851: hmm that's possible - I like that! Let me look into it!

ben851: Yup - that works! Thanks for the idea!

PR #2723, opened 2024-07-02
Revert "turn off default pool in staging"

sastels:
Reverts cds-snc/notification-manifests#2704


sastels:
conflict, will do manually

PR #2734, opened 2024-07-04
[MANIFEST] - Adding vpn back

ben851:
The 1Password service account has been created with the right permissions. Adding back VPN

## What are you changing?
- [x] Changing github workflows


sastels (approved):
LGTM

PR #2828, opened 2024-08-01
Adding Xray Env Var

P0NDER0SA:
## What are you changing?
Adding an enviroment variable

Granting doc download api rights to point to the xray daemonset

### DIFF
--- a/env/dev/kustomization.yaml
+++ b/env/dev/kustomization.yaml
@@ -677,10 +677,3 @@ vars:
     apiVersion: v1
     kind: ConfigMap
     name: application-config
-- fieldref:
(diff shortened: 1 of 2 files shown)


env/dev/kustomization.yaml
```diff
@@ -677,10 +677,3 @@ vars:
     apiVersion: v1
     kind: ConfigMap
     name: application-config
-- fieldref:
```
sastels, line 680: You meant to take out XRAY for dev?

sastels (approved):
one question

env/dev/kustomization.yaml
P0NDER0SA: ya, that was old code that had no business being there.

PR #2791, opened 2024-07-24
Paring down kube state metrics

ben851:
## What happens when your PR merges?
Attempting to pare done the number of metrics sent by KSM.

## What are you changing?
- [ ] Releasing a new version of Notify
- [x] Changing kubernetes configuration

## Provide some background on the changes
Investigating New Relic costing, and trying to reduce cloudwatch costing

## Checklist if making changes to Kubernetes:
- [x] I know how to get kubectl credentials in case it catches on fire

## After merging this PR
- [ ] I have verified that the tests / deployment actions succeeded
- [ ] I have verified that any affected pods were restarted successfully
- [ ] I have verified that I can still log into [Notify production](https://notification.canada.ca)
- [ ] I have verified that the smoke tests still pass on production
- [ ] I have communicated the release in the #notify Slack channel.

PR #2830, opened 2024-08-01
Revert "Paring down kube state metrics"

sastels:
Reverts cds-snc/notification-manifests#2791

Let's see how much of Ben's work we can revert before he comes back. And this one might even fix the missing data in our dashboards!


jimleroyer (approved):

P0NDER0SA (approved):
He's going to LOVE IT!

PR #2842, opened 2024-08-06
fix: fix production helm plan

sastels:

## What are you changing?

The production helm plan has always been broken but we don't usually trigger it (ie with a direct VERSION change) so we haven't noticed until now!


sastels:
ok fixed action was run here
https://github.com/cds-snc/notification-manifests/actions/runs/10269942930/job/28416566152
(then VERSION was reverted)

P0NDER0SA (approved):
Yep! As discussed

PR #2853, opened 2024-08-12
[MANIFEST] Switch on FF for terms of use in production

jzbahrai:
```
➜  notification-manifests git:(task/terms-use) ✗ make diff-production
Archive:  .previous.env.zip
  inflating: .env
Archive:  .env.zip
  inflating: .env
48c48
< FF_TOU=false
---
> FF_TOU=true
```


whabanks (approved):
LGTM!
```
make diff-production
Archive:  .previous.env.zip
  inflating: .env
Archive:  .env.zip
  inflating: .env
48c48
< FF_TOU=false
---
> FF_TOU=true
```

PR #2858, opened 2024-08-14
[MANIFEST] Added 3 new sensitive services

jzbahrai:
Will remove the TWILIO keys in another PR:

Please run make diff-production to see the diff


whabanks (approved):
LGTM! 3 new services are present in the `SENSITIVE_SERVICES` key

jimleroyer:
Can we apply the same changes to the twilio env vars for the staging file as well? Hence we will see in staging if that causes any issue (it should not, but in case).

PR #2918, opened 2024-09-05
Fixes for Dev

ben851:
Dev only - I have hard-coded the target type in the target group bindings, and removed the aws-auth config map which is now managed by terraform for dev.

## What are you changing?

Dev was failing creation due to bad aws-auth configmap, and target group bindings being created too quickly.

## After merging this PR


P0NDER0SA (approved):
This works!

PR #2981, opened 2024-09-23
set staging to new sms config

sastels:
## What are you changing?

Set staging to a new sms config that should get us closer to 6000 fragments / min.
- max 15 send-sms-scalable pods
- 10/s deliver_sms rate limit

```
$ make diff-staging
< CELERY_DELIVER_SMS_RATE_LIMIT=2/s
---
> CELERY_DELIVER_SMS_RATE_LIMIT=10/s
```


P0NDER0SA (approved):
Let's do this!!

PR #2985, opened 2024-09-24
set staging sms hpa to max 27

sastels:
## What are you changing?

Set the `celery-sms-send-scalable-hpa` max pods to 27. This will give us 30 sms sending pods in total, which should send at a rate of around 5000 notifications per minute (which would be 10000 fragments / minute for our average of 2 fragments / sms in production).

We'll use this to test (using internal test numbers) what happens to the system if we try to send at this rate! 📈


ben851 (approved):
YEAHHHH

PR #2990, opened 2024-09-26
change staging sms hpa to 4

sastels:
## What are you changing?

Change the staging sms-send-scalable hpa to ~4~ 3 max pods. This should give us around 2200 notifications / minute, ie about 4400 fragments / minute in prod (2 fragments / sms average), matching the send rate of our current 74 long codes.

Tested in dev and got about 2700 npm which will probably be fine if we're also doing the AWS network calls.


ben851 (approved):
yes

sastels:
Gonna test in dev first, might do this too for staging but maybe not. Moving back to draft for now...

PR #3003, opened 2024-10-01
[MANIFEST] Set sms performance parameters for prod to send about 4400 fragments per minute

sastels:

## What are you changing?

- change task rate limit to 10/s
- lower sms hpa to max 3 pods

```
$ make diff-production
< CELERY_DELIVER_SMS_RATE_LIMIT=1/s
---
> CELERY_DELIVER_SMS_RATE_LIMIT=10/s
```

Set parameters around sms sending to send at a rate of about 2200 notifications per minute, which will be (with our average 2 fragments / sms) about 4400 fragments per minute. This is the capability of our current 74 production long codes.

Note that if we go over we will
- get a throttling warning in #notification-ops
- get retried 5 minutes later. The retry will be on SNS and will use our short code if needed, so will succeed.

If we see a lot of throttling warnings then we should tweak our rate a bit lower. We could also revert this PR to get us back to the old configuration.

## After merging this PR


ben851 (approved):
LGTM

PR #3014, opened 2024-10-02
Revert "[MANIFEST] Set sms performance parameters for prod to send about 4400 fragments per minute"

sastels:
Reverts cds-snc/notification-manifests#3003

Just merge this if we get lots of `ThrottlingExceptions`!

Note that if someone changes the encrypted prod variables then we'll get a merge conflict. You'll have to set the `CELERY_DELIVER_SMS_RATE_LIMIT` to 1/s


sastels:
Dont need

PR #3083, opened 2024-11-07
Helmfile deploying on dev - notify-documentation

P0NDER0SA:
## What are you changing?

Updating the helmfile code to deploy the documentation portion.  No secrets are required, so only a few code updates were required.

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/450

## After merging this PR


helmfile/charts/notify-documentation/templates/deployment.yaml
```diff
@@ -3,22 +3,22 @@ kind: Deployment
 metadata:
   name: {{ include "notify-documentation.fullname" . }}
   labels:
-    {{- include "notify-documentation.labels" . | nindent 4 }}
+    app: {{ include "notify-documentation.name" . }}
```
ben851, line 6: Being pedantic again, but this should probably be in camel case like everything else... notifyDocumentation... or just documentation really since we don't say notifyApi or notifyAdmin in the others.

PR #3099, opened 2024-11-14
Adding startup probe on the API pods

jimleroyer:

  - `BREAKING CHANGE:` - tag `main` as a new major release
  - `chore:` - use for changes to non-app code (ex: GitHub actions)
- Alternatively, change the [VERSION file](https://github.com/cds-snc/notification-manifests/blob/main/VERSION) - this will not create a new tag, but rather will release the tag in `VERSION` to production.

## What are you changing?

Added a startup probe on the API so that there is more initialization time prior for the readiness and live probes to take effect. We need more init time because of recent changes introduced in the New Relic agent library v8.10.1 which introduced richer retrieval of dependencies metadata. The probe configuration previous to these changes triggered a potential loop crash effect because the API workers would get restarted constantly and trigger the New Relic more time consuming process **everytime**.

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document download API

## After merging this PR


base/notify-api/api-deployment.yaml
```diff
@@ -283,20 +283,29 @@ spec:
...
             - containerPort: 6011
+          startupProbe:
+            httpGet:
+              path: /_status?simple=true
+              port: 6011
+            initialDelaySeconds: 15
+            periodSeconds: 3
+            timeoutSeconds: 1
+            successThreshold: 1
+            failureThreshold: 6
```
jimleroyer, line 294: The startup probe will make its first check after 15 seconds, after which it will try 6 times every 3 seconds, tagging the startup as a definitive failure after 33 seconds (15 + 6 * 3). 

An interesting issue we have with the slower New Relic initialization is that the latter is only triggered after the first request hits the API. That means that while the start up probe might be successful after the first request and then let the readiness and live probes to handle the rest, it might be misleading as New Relic would lock the app during its initialization.

Hence it is quite possible that we need to increase the `initialDelaySconds` of 15 seconds here.

Also note that new changes in gunicorn with regard to its timeout configuration will be as follows:

1. `gracefulTimeout` set to 25 seconds; in order to have the New Relic initialization roll out properly before gunicorn restarts the worker and in return, make the k8s pod reaches its own timeout.
2. `timeout` set to 30 seconds; this is the hard timeout for gunicorn to let its workers initialized.

Hence the 33 seconds failure for the k8s startupProbe is aligned with the hard timeout of gunicorn. That said, given the first request might be successful per reasons described above but make the rest of the application main thread locked, we might have to tweak this further.

base/notify-api/api-deployment.yaml
```diff
@@ -283,20 +283,29 @@ spec:
...
             timeoutSeconds: 1
             successThreshold: 3
-            failureThreshold: 10
+            failureThreshold: 5
           livenessProbe:
             httpGet:
               path: "/_status?simple=true"
               port: 6011
-            initialDelaySeconds: 30
+            initialDelaySeconds: 5
```
jimleroyer, line 308: These values above should be lower now that we have a startupProbe, waiting on proper initialization of the application.

ben851 (approved):
🆒  Nice!

base/notify-api/api-deployment.yaml
P0NDER0SA: Nice description!

let’s try this

P0NDER0SA (approved):

PR #3109, opened 2024-11-15
Tweak the API probes to accommodate xray and newrelic init

jimleroyer:

The API probes should be more reliable for the API pods deployment. 

## What are you changing?

With the recent upgrade of New Relic to version 9.2.0, we reviewed the existing probes (startup, liveness and readiness) for the API, to take in account initialization of the former and of AWS X-Ray. As the startup time of our application can get longer, we need to accommodate for these. 

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document download API

## After merging this PR


P0NDER0SA (approved):
This is good!

base/notify-api/api-deployment.yaml
```diff
@@ -287,11 +287,11 @@ spec:
             httpGet:
               path: /_status?simple=true
               port: 6011
-            initialDelaySeconds: 15
+            initialDelaySeconds: 20
             periodSeconds: 3
             timeoutSeconds: 1
             successThreshold: 1
-            failureThreshold: 6
+            failureThreshold: 20
```
jimleroyer, line 294: We are total giving 1m20s of start up time to the app, with 20 seconds being the minimum initialization and then trying to get the ball rolling with a http test every 3 seconds, for 1 minute.

base/notify-api/api-deployment.yaml
```diff
@@ -300,12 +300,12 @@ spec:
             periodSeconds: 3
             timeoutSeconds: 1
             successThreshold: 3
-            failureThreshold: 5
+            failureThreshold: 10
```
jimleroyer, line 303: Once the startup probe is done, the readiness and liveness probes will kick in. The readiness will determine when the app is ready and might need to wait more time than what the startup probe allowed. The reason for that is that the first http hit will trigger an initialization for lazy loaded components such as New Relic. So we need to give it more flexibility. 

A failure threshold of 10 for 3 period seconds means that a total of 30 seconds will be tried, but after waiting for an initial delay seconds of 5 seconds. Hence 35 seconds is allowed for this readiness stage.

base/notify-api/api-deployment.yaml
P0NDER0SA: Interesting!

PR #3121, opened 2024-11-21
smoke test secrets removed

ben851:

Removing the smoke test secrets since they aren't secret. Also switching the smoke access key and access key secret to be the normal ones for staging. They would have been this anyway since there is only one access key/secret in staging AWS.

## What are you changing?

- [x] Github Config

In preparation for cleaning up github secrets.

## After merging this PR


P0NDER0SA (approved):
Perfect!

PR #3130, opened 2024-11-21
Removing non specific OP SA Token

ben851:

Switch to using the OP SA token secret that has the environment name in it

Also cleaning up kubeconfig code to use aliases instead of context renaming

Also Also normalizing the staging secret for aws account id

## After merging this PR


P0NDER0SA (approved):
Yep

PR #3136, opened 2024-11-26
[MANIFEST] clear cache fix

ben851:

This will fix the invalid workflow config on merge to main prod


sastels (approved):
ah right, these aren't secrets anymore

PR #3153, opened 2024-11-28
Finalizing helm overrides

P0NDER0SA:
Nothing until we turn these on for staging and production.  It can be used to deploy on DEV.

## What are you changing?
Updating the overrides for our Helm migration. They are now "complete"

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/457

## After merging this PR

### DIFF
--- a/helmfile/charts/notify-celery/templates/pdb.yaml
+++ b/helmfile/charts/notify-celery/templates/pdb.yaml
@@ -1,13 +1,13 @@
 {{- range $node := .Values.nodes }}
 {{- with $ -}}
-{{ if $node.pdbEnabled }}
+{{ if $node.pdb.pdbEnabled }}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: notify-celery-{{ $node.name }}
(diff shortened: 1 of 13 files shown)


helmfile/charts/notify-celery/templates/pdb.yaml
```diff
@@ -1,13 +1,13 @@
...
-{{ if $node.pdbEnabled }}
+{{ if $node.pdb.pdbEnabled }}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: notify-celery-{{ $node.name }}
   namespace: {{ .Release.Namespace }}
 spec:
-  minAvailable: {{ .Values.replicaCount }}
+  minAvailable: {{ $node.pdb.minAvailable }}
```
ben851, line 10: good catch here - if we used replica count we'd never be able to rollover

ben851 (approved):
Approving to get this going since it's dev only

PR #3193, opened 2024-12-12
adding annual limit config for helm

P0NDER0SA:

Adds a configuration for FF_ANNUAL_LIMIT that was missed because it was added after we started migrating our code.

## What are you changing?

Just a configuration variable

This needs to be added for our helm deployments now that they are officially on staging.

## After merging this PR


ben851 (approved):
LGTM

PR #3287, opened 2025-01-08
Workflow to update Docker Image Tags for Components

P0NDER0SA:

Nothing until the workflow gets triggered by outside repositories (ideally api, admin, dd and documentation)

## What are you changing?

Adding a workflow that's triggered only on dispatch

We need our component repositories to trigger updates to our docker image tags in our .env files on the main branch.  This workflow will accept those tags and update the files.

## After merging this PR


ben851 (approved):
LGTM

PR #3342, opened 2025-01-15
scaling beats to 0 for celery

ben851:

Scale Prod beat to 0 in helmfile while we are testing helmfile. 

## What are you changing?

We had an incident last night about invalid tables that could be caused by having two celery beats running.

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document download API
- [x] Celery

## After merging this PR

### DIFF
--- a/helmfile/overrides/notify/celery.yaml.gotmpl
+++ b/helmfile/overrides/notify/celery.yaml.gotmpl
@@ -43,7 +43,7 @@ nodes:
     deployment:
       deploymentBeatEnabled: {{ if eq .Environment.Name "production" }} true {{ else if eq .Environment.Name "staging" }} true {{ else }} true {{ end }}
       terminationGracePeriodSeconds: {{ if eq .Environment.Name "production" }} 30 {{ else if eq .Environment.Name "staging" }} 30 {{ else }} 30 {{ end }}
-      replicas: {{ if eq .Environment.Name "production" }} 3 {{ else if eq .Environment.Name "staging" }} 3 {{ else }} 3 {{ end }}
+      replicas: {{ if eq .Environment.Name "production" }} 0 {{ else if eq .Environment.Name "staging" }} 3 {{ else }} 3 {{ end }}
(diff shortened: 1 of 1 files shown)


helmfile/overrides/notify/celery.yaml.gotmpl
```diff
@@ -43,7 +43,7 @@ nodes:
     deployment:
       deploymentBeatEnabled: {{ if eq .Environment.Name "production" }} true {{ else if eq .Environment.Name "staging" }} true {{ else }} true {{ end }}
       terminationGracePeriodSeconds: {{ if eq .Environment.Name "production" }} 30 {{ else if eq .Environment.Name "staging" }} 30 {{ else }} 30 {{ end }}
-      replicas: {{ if eq .Environment.Name "production" }} 3 {{ else if eq .Environment.Name "staging" }} 3 {{ else }} 3 {{ end }}
+      replicas: {{ if eq .Environment.Name "production" }} 0 {{ else if eq .Environment.Name "staging" }} 3 {{ else }} 3 {{ end }}
```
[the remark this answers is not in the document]
ben851: Just the beat

helmfile/overrides/notify/celery.yaml.gotmpl
```diff
@@ -43,7 +43,7 @@ nodes:
     deployment:
       deploymentBeatEnabled: {{ if eq .Environment.Name "production" }} true {{ else if eq .Environment.Name "staging" }} true {{ else }} true {{ end }}
```
jimleroyer, line 44: Could we put the value to false here? different way and same result?

jimleroyer (approved):

helmfile/overrides/notify/celery.yaml.gotmpl
ben851: Yes, with this way the deployment won't deploy... which would not validate the code... I'd rather deploy with 0 to validate that the code at least works.

PR #3350, opened 2025-01-16
Setting the correct launch for sms

ben851:

The new helm celery sms send for other-celery (dedicated long code sms sending) was using the celery-beat shell script for launch, which is incorrect and was causing an issue in prod.

## What are you changing?

Helmfile migration

## After merging this PR

### DIFF
--- a/helmfile/charts/notify-celery/values.yaml
+++ b/helmfile/charts/notify-celery/values.yaml
@@ -47,23 +47,7 @@ celerySecrets:
   POSTGRES_HOST: MANIFEST_POSTGRES_HOST
 
 nodes:
-  # BEAT CONFIGS
-  other-sms:
-    name: sms
-    newRelicAppName: "notification-celery-sms-dev"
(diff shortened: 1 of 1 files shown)


sastels (approved):
LGTM. Thanks for adding the comments.

helmfile/charts/notify-celery/values.yaml
```diff
@@ -156,7 +143,25 @@ nodes:
       hpaEnabled: false
     service:
       serviceEnabled: true
+  # CELERY FOR SENDING SMS OVER DEDICATED LONG CODE
+  other-sms:
+    name: sms
+    newRelicAppName: "notification-celery-sms-dev"
+    scriptArg: "sh /app/scripts/run_celery_sms.sh"
```
jimleroyer, line 150: we should probably rename that script too, so it's obvious what the usage is cc @sastels

sastels: yes a bit of renaming with "dedicated" or "throttled" in there would certainly help.

PR #3361, opened 2025-01-20
fix: log if database migration fails

sastels:

## What are you changing?

Explicitly log if the database migration `flask db upgrade` fails.

### DIFF
--- a/base/notify-api/api-deployment.yaml
+++ b/base/notify-api/api-deployment.yaml
@@ -135,7 +135,7 @@ spec:
             [
               "sh",
               "-c",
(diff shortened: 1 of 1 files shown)


base/notify-api/api-deployment.yaml
```diff
@@ -135,7 +135,7 @@ spec:
             [
               "sh",
               "-c",
-              "flask db upgrade",
+              "flask db upgrade || echo 'ERROR: database migration failed'",
```
jimleroyer, line 138: Nice trick <3

PR #3367, opened 2025-01-20
fix: increase smoke test timeout

sastels:

## What are you changing?

Smoke test is timing out waiting for delivery receipts to be retried.
Here we bump up the timeout to larger than what would have been needed in the latest timeout.


ben851 (approved):
LGTM

PR #3371, opened 2025-01-21
Removing old variables test

P0NDER0SA:

Nothing - we have removed a workflow test that is no longer required.

## What are you changing?

removing a workflow file

We've migrated to helmfile and have changed completely how we manage our environment variables.  Because of this, we no longer need to check whether our env files are up to date.

## After merging this PR


ben851 (approved):
LGTM

PR #3389, opened 2025-01-22
[MANIFEST] Upping celery memory requests to 800

ben851:

We're seeing pod eviction due to running out of memory on staging Nodes. Lets up the request to something higher so that K8s doesn't try and put too many celery pods on the same nodes.

## What are you changing?

Incident debugging

## After merging this PR


P0NDER0SA (approved):
Looks good

PR #3398, opened 2025-01-23
[MANIFEST] Helmfile Prod Plan Fix

ben851:

Debugging helmfile prod diff - we're getting a syntax error on this line... For some reason even though this line looks identical to me, it's showing it as a change so want to see whats up.

## After merging this PR


P0NDER0SA (approved):
LGTM

PR #3402, opened 2025-01-23
remove if condition on staging helmfile diff

ben851:

We'll need to run staging diff even on prod-prs for the time being. I have an idea on how to fix this properly but it will require an overhaul of the workflows.

## What are you changing?

- [x] Github workflows

## After merging this PR


P0NDER0SA (approved):
lgtm

PR #3410, opened 2025-01-27
Prep for helmfile

ben851:

Removing all the old stuff from kustomize for dev - will still need to migrate the system stuff but we haven't done that anywhere yet.

## What are you changing?

Helmfile migraiton

## After merging this PR


P0NDER0SA (approved):
excellent, this will make things clearer for our system migration.  Perfect

PR #3437, opened 2025-01-27
Migrating ipv4 service/deployment to Helmfile

P0NDER0SA:

IPV4 deployment and Service in the notification-canada-ca namespace will be migrated to helm.  Functionality should remain the same.  ipv4 service and deployment (and the urls that use it) are all updated to ipv4-geolocate.

## What are you changing?

Migrating the ipv4 code from Kustomize to Helmfile/Helm

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/466

## After merging this PR


helmfile/charts/ipv4/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "ipv4.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
```
jimleroyer, line 5: This is some kind of naming standardization step?

helmfile/charts/ipv4/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
```
jimleroyer, line 11: Ah here is the explanation from my question above I guess.

P0NDER0SA: ya, so these are built automatically during the bootstrapping of a helm project (helm create).

ben851 (approved):
Approved but can we rename the chart and deployment name to ipv4-geolocate so we have a better idea of what it is in the future?

P0NDER0SA:
> Approved but can we rename the chart and deployment name to ipv4-geolocate so we have a better idea of what it is in the future?

I agree... this might actually help us avoid the issue of overwriting deployments 🤔 we should try and rename all of these to make them more descriptive.

helmfile/charts/notify-admin/values.yaml
```diff
@@ -10,7 +10,7 @@ admin:
   ASSET_UPLOAD_BUCKET_NAME: "notification-canada-ca-default-asset-upload"
   BULK_SEND_AWS_BUCKET: "notification-canada-ca-default-bulk-send"
   CSV_UPLOAD_BUCKET_NAME: "notification-canada-ca-default-csv"
-  IP_GEOLOCATE_SERVICE: "http://ipv4.notification-canada-ca.svc.cluster.local:8080"
+  IP_GEOLOCATE_SERVICE: "http://ipv4-geolocate.notification-canada-ca.svc.cluster.local:8080"
```
ben851, line 13: Good catch 👍

P0NDER0SA: i know 🧠

PR #3470, opened 2025-01-29
[MANIFEST] Removing the default since it seems to not be working as expected

ben851:

The default latest seems to be behaving weirdly. Removing that.

## What are you changing?

## After merging this PR

### DIFF
--- a/helmfile/overrides/notify/admin.yaml.gotmpl
+++ b/helmfile/overrides/notify/admin.yaml.gotmpl
@@ -29,7 +29,7 @@ admin:
   GC_ORGANISATIONS_BUCKET_NAME: "{{ .StateValues.GC_ORGANISATIONS_BUCKET_NAME }}"
 
 image:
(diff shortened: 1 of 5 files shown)


P0NDER0SA (approved):
weird, but yes -- this fixes it

helmfile/overrides/notify/admin.yaml.gotmpl
```diff
@@ -29,7 +29,7 @@ admin:
   GC_ORGANISATIONS_BUCKET_NAME: "{{ .StateValues.GC_ORGANISATIONS_BUCKET_NAME }}"
 
 image:
-  tag: {{ .StateValues.ADMIN_DOCKER_TAG | default "latest" }}
+  tag: "{{ .StateValues.ADMIN_DOCKER_TAG }}"
```
sastels, line 32: Quotes here but not in others?

PR #3473, opened 2025-01-29
Helm fluentbit agent

P0NDER0SA:

fluentbit gets deployed as fb-agent-fluent-bit using helm

## What are you changing?

Updating and migrating fluentbit to be deployed with Helm

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/507

## Release Steps

- Delete the old Kustomize deployment before merging
- Pull latest code from main on notification-manifests
- navigate to 'env/staging' folder
- run `kubectl delete -f fluentbit.yaml   `
- merge PR 

## After merging this PR


helmfile/overrides/system/fluentbit-agent.yaml.gotmpl
```diff
@@ -0,0 +1,401 @@
+env:
```
jimleroyer, line 1: I remember we had some difficulties with fluentbit back when we enabled it with @ben851 using kustomize. Is that a copy of the same configuration? 👀

P0NDER0SA: I'll take this up on a. slack DM chat, but the quick answer is that it's the same configurations, but they've been completely migrated and modified where necessary to be a helmfile overrides file

ben851 (approved):
We should probably delete the old one first

PR #3516, opened 2025-02-05
adding key that was missing (it was being set in defaults with TF)

P0NDER0SA:

missing a key that wasn't there

## After merging this PR


ben851 (approved):
LGTM

PR #3570, opened 2025-02-18
Cache fix

ben851:

Upgrade the actions cache to latest

## What are you changing?

- [x] Github config

OId cache is out of date

## After merging this PR


.github/workflows/smoke_test.yaml
```diff
@@ -31,13 +31,12 @@ jobs:
...
         run: python -m pip install --upgrade pip
 
-      - name: Get packages from cache
-        uses: actions/cache@e12d46a63a90f2fae62d114769bbf2a179198b5c # v3.3.3
+      - uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
         with:
-          path: ~/.local
-          key: poetry-${{ runner.os }}-${{ env.PYTHON_VERSION }}-${{ hashFiles('**/poetry.lock') }}
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```
sastels, line 37: wait, where is this `requirements.txt` file?

sastels: ah this is in the smoke test action. So that's using the api repo, which is poetry, so maybe it should be using the poetry.lock file from that repo for the hash key?

PR #3581, opened 2025-02-19
[MANIFEST] FIX Adding FIDO2_DOMAIN Var to API

P0NDER0SA:

Adding FIDO2_DOMAIN variable to the API. This will fix some errors we have with yubikey and passkey for 2fa in production

## What are you changing?

FIDO2_DOMAIN: "dev.notification.cdssandbox.xyz" (setting this as required by environment)

https://docs.google.com/document/d/1055vlAcC2sXZCka1ogH7zU9QeJzASUhQuWJ_yGd3WPk/edit

## After merging this PR


P0NDER0SA:
> Where do the values for staging/prod come from?

they are found in the production.env and staging.env files in the overrides folder.... the 

> Where do the values for staging/prod come from?

in that particular instance they are pulled from the context variables that are in the process.  We could define them in the staging and production .env files in overrides as well -- either will get the job done.  Let me know if you need more info :)

PR #3588, opened 2025-02-19
Updating Notify Components Helm Overrides

P0NDER0SA:
## What are you changing?

I'm moving the secrets and variables for api to a different file (overrides) so that there is less redundant declarations and the code is easier to understand.

Helm should show no changes as this is meant to work exactly like it did before, but with cleaner code.

this improvement came from the following incident:
https://docs.google.com/document/d/1055vlAcC2sXZCka1ogH7zU9QeJzASUhQuWJ_yGd3WPk/edit

## After merging this PR


P0NDER0SA:
No DIFF! That's great 🥇

ben851 (approved):
LGTM

PR #3613, opened 2025-02-26
chore: env var consistency check

sastels:

Add a GitHub action and workflow to compare the helm environment variable overrides to ensure that the files contain the same variables.

Restored the [old code we had for comparing the encrypted env vars ](https://github.com/cds-snc/notification-manifests/blob/75f55d77e3c741a443ed1c2b05b5bf3bd12c5c63/.github/workflows/test_env_vars.yaml)and tweaked to apply to the new unencrypted helm overrides.

Tested by adding `STEVE_RETIRED: "true"` to dev. Workflow fails with

image


ben851 (approved):
LGTM

P0NDER0SA (approved):
Nice! :)

PR #3614, opened 2025-02-26
increase sms sending max pods on staging and dev

sastels:
## What are you changing?

The sms send rate is lower than our AWS resources will support. This scales it up a bit on staging and dev.


ben851 (approved):
LGTM

PR #3616, opened 2025-02-27
fix: set staging and dev to one beat worker

sastels:

## What are you changing?

We set dev and staging to have three beat workers when we switched to helm. This may explain many of the celery errors we've been seeing in staging...


P0NDER0SA (approved):
perfect 👍🏻

PR #3620, opened 2025-02-27
celery multiline

ben851:

This will fix (for the most part) celery multiline issues. There are a couple outstanding oneoffs, but the main thing is that stack traces work now. 

## What are you changing?

Celery multiline on fluentbit

## After merging this PR


P0NDER0SA (approved):
computing regex.... computing .... computing....

approved 🔔

PR #3644, opened 2025-03-11
set staging and dev sms-send-scalable hpa back to max 3 pods

sastels:
## What are you changing?

We set the sms-send hpa to go up to 6 pods, here we dial it back to 3 (same as prod and same as the original value speced out in September)


ben851 (approved):
LGTM

PR #3678, opened 2025-03-20
Workflow trigger fix

P0NDER0SA:

It should run DB upgrades as a part of our Helmfile Apply workflow -- but the subsequent steps will not run unless the DB upgrade works successfully

## What are you changing?

Adding an action for the DB upgrade and calling it from the necessary helmfile apply workflow (for staging)

This is a modification to have this work a little more streamlined and make our helmfile apply easily dependent on the db upgrade working

## After merging this PR


ben851 (approved):
LGTM

PR #3692, opened 2025-03-25
Removing old DB Upgrade from Production API Pod

P0NDER0SA:

The K8s API Pod will no longer have the DB upgrade code run during its deployment.  This is now done in a separate step/pod prior to deploying the API.  The Lambda API and K8s API  deployments will both now be independent of the API K8s deployment.

## What are you changing?

Updating the API K8s deployment (Helmfile) file to remove the DB upgrade scripts.

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/20

## After merging this PR


ben851 (approved):
LGTM

PR #3693, opened 2025-03-26
Adding the Revision Number to our DB Upgrade Job

P0NDER0SA:

This will just add the revision number to our DB Upgrade jobs for better identification, and ensuring that it'll run every time, even if the same tag is being used.

## What are you changing?

Minor update to the DB upgrade job and DB Upgrade GHA

This is related to our work on removing the DB upgrade from our API K8s deployment

## After merging this PR


ben851 (approved):
neato

jimleroyer (approved):

PR #3699, opened 2025-03-31
adding step label to db upgrade job for new env creation

ben851:

We need the database upgrade job to run before API fires up on new environment creation. Setting the helmfile label as step 4 so that it gets run as a pre-req.

## What are you changing?

Dev env creation

## After merging this PR


jimleroyer (approved):
LGTM

PR #3701, opened 2025-04-01
adding node selectors to system critical apps

ben851:

Trying to investigate the 504 gateway timeouts. I can sometimes correlate a node being recycled around the time the 504 appears. When we switched from Kustomize to Helmfile we didn't always add the node selector for system critical apps to be ON_DEMAND. Adding that here. Not 100% sure this will fix anything but it should be there anyway.

## What are you changing?

Investigating 504 errors in prod

## After merging this PR


P0NDER0SA (approved):
LGTM

jimleroyer (approved):
Thanks for looking into this.

PR #3702, opened 2025-04-01
adding workflows to run latest db upgrade manually

P0NDER0SA:

Nothing, this will only run if manually triggered

## What are you changing?

Updates to our Github Workflows (adding two) for Prod and Staging to run the latest API tag DB upgrade script.

## After merging this PR


jimleroyer (approved):
Looks good :shipit:

PR #3751, opened 2025-04-17
Adding Password and reducing code redundancy for Flask DB Workflows

P0NDER0SA:
Nothing until we run the workflow.

The workflows should accept a password stored in 1pass (for the production workflow) and have dropdowns to select upgrade or downgrade when manuallly running GHA job.

## What are you changing?

Reducing workflows and adding some extra security around manual DB workflows (via password)

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/558

## After merging this PR


.github/workflows/db_production_flask_migration_manual.yaml
```diff
@@ -1,7 +1,19 @@
-name: Production - Manual DB Downgrade
+name: Production - Manual DB Flask Command
 
 on:
   workflow_dispatch:
+    inputs:
+      password:
+        description: "Enter the password - can be found here: https://start.1password.com/open/i?a=BZ4RYZZICJEKNOF4Z6ASRX2XXI&v=ppnxsriom3alsxj4ogikyjxlzi&i=i5kfypu7grhf5lc2bu4mi4yque&h=cds-snc.1password.ca"
```
jimleroyer, line 7: Can we put in a link in here? markdown format?

.github/workflows/db_production_flask_migration_manual.yaml
```diff
@@ -1,7 +1,19 @@
...
+        required: true
+        default: ""
+      flask_command:
+        description: "Select the Flask command to run"
+        required: true
+        type: choice
+        options:
+          - upgrade
+          - downgrade
```
jimleroyer, line 16: That should be sufficient for us at this point in time 👍 If we need to pass in versions, that can be another iterative improvement.

.github/workflows/db_production_flask_migration_manual.yaml
P0NDER0SA: I had the same thought -- quickly couldn't find a way, but I'll look deeper

jimleroyer (approved):
lovely <3

PR #3786, opened 2025-04-30
docker hub authentication

ben851:

We will use authentication with docker hub to increase our pull limit. 

## What are you changing?

We are hitting rate limits in staging on docker hub pulls.

## After merging this PR


helmfile/getContext.ps1
jimleroyer, line 1: We can thrash this file.

helmfile/charts/notify-celery/values.yaml
```diff
@@ -323,3 +323,7 @@ secretProviderClass:
           objectName: INTERNAL_DNS_CERT_BASE64
       secretName: nginx
       type: kubernetes.io/tls
+
+dockerhub:
+  username: "changeme"
+  pat: "changeme"
```
jimleroyer, line 329: We won't change @patheard

patheard: I respect it

PR #3791, opened 2025-05-01
Docker Image Pull Secrets Fix

ben851:

I had the docker image pull secret in the wrong place and also screwed up the naming. All in all,  a complete failure on my part. I've verified this fixes it.

## What are you changing?

Docker 429 rate limit

## After merging this PR

### DIFF
--- a/helmfile/overrides/notify/admin.yaml.gotmpl
+++ b/helmfile/overrides/notify/admin.yaml.gotmpl
@@ -64,7 +64,7 @@ image:
 
 dockerhub:
   username: {{ requiredEnv "MANIFEST_DOCKER_HUB_USERNAME" }}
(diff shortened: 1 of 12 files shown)


helmfile/overrides/notify/admin.yaml.gotmpl
```diff
@@ -64,7 +64,7 @@ image:
 
 dockerhub:
   username: {{ requiredEnv "MANIFEST_DOCKER_HUB_USERNAME" }}
-  password: {{ requiredEnv "MANIFEST_DOCKER_HUB_PAT" }}
+  pat: {{ requiredEnv "MANIFEST_DOCKER_HUB_PAT" }}
```
jimleroyer, line 67: I should have caught this one..

jimleroyer (approved):
Looks better to me!

PR #3797, opened 2025-05-02
Removing the custom crafted github action to test the pw protected gh action workflow

jimleroyer:
Reverts cds-snc/notification-manifests#3793

This was only needed to test the password protected github action. The original one is operating on production and we didn't want to test in production. The code was the same except targeting the staging environment.


P0NDER0SA (approved):
Hasta la Vista file! 💥

PR #3804, opened 2025-05-06
Disabling Celery Reports PDB

ben851:

Disabling the celery reports pod disruption budget since we only have one replica. We need this to be disabled so that we can upgrade k8s nodes.

## What are you changing?

## After merging this PR


P0NDER0SA (approved):
LGTM

PR #3831, opened 2025-05-14
Image Pull Secrets for K8s

ben851:

Image Pull secrets are bound to a namespace so we need one for other namespaces.

## What are you changing?

Stupid docker rate limit issues

## After merging this PR


P0NDER0SA (approved):
LGTM

helmfile/charts/image-pull-secrets/templates/_helpers.tpl
jimleroyer, line 1: Is there a formatter for these? to have the conditions stand out and its content indented?

ben851: I'm not sure - this is straight from helm so I would assume if there was one, it would have done it...

jimleroyer: Alright

PR #4484, opened 2026-02-09
put staging back to known good

ben851:

I tried lowering the staging k8s api resources to what's in line with prod, but that doesn't seem to be enough for perf tests. I'm going to put the higher numbers back and leave it like that for a while. Let's see if it settles down.

## What are you changing?

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/770

## After merging this PR


P0NDER0SA (approved):
looks good

PR #4740, opened 2026-04-27
Api hpa tuning

ben851:

API Pods will scale on 150% of request CPU usage (approx 75% of limit usage). When scaling in staging, there is a 10 minute stabilization window, and it will only scale down 1 pod per 5 minutes so that we can be sure that the load is consistent.

## What are you changing?

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/761

## After merging this PR


jimleroyer (approved):
Very nice

PR #4750, opened 2026-04-29
Fix API node affinitiy

ben851:

The spot API instances were inheriting the node selector from the non spot which caused a conflict. Fixed here.

Redline testing

## After merging this PR


P0NDER0SA (approved):
OK

PR #4784, opened 2026-05-15
[MANIFEST] Fix signoz

ben851:

We didn't have the OTLP versions pinned for signoz stuff and a chart upgrade messed everything up. I've pinned all of the charts - even the ones that don't need pinning because they're tied to our own charts in the repo just to make it consistent.

## What are you changing?

Signoz done broke

## After merging this PR


jimleroyer (approved):
Yes please, we need to pin every version for security purposes anyway.

PR #4785, opened 2026-05-15
need to add priority class to otel agent

ben851:

The otel agent isn't inheriting the priority class from the root. Adding it specifically here so that it always preempts other pods.

## What are you changing?

Came across this issue while investigating prod signoz

## After merging this PR


cds-stephenchow (approved):
ok

PR #4791, opened 2026-05-20
migrating to new aws secrets provider

ben851:

Theoretically this will delete the old aws csi driver, and then install the updated version along with the updated ascp helm chart. The csi driver is now bundled in this chart. 

If this doesn't work, some manual steps will be required at release time, but I wanted to see if I could get the automation to do it all first.

## What are you changing?

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/852

## After merging this PR


helmfile/helmfile.yaml.gotmpl
```diff
@@ -226,13 +227,22 @@ releases:
 
   - name: aws-secrets-provider
     namespace: kube-system
+    needs:
+      - kube-system/secrets-store-csi-driver
```
jimleroyer, line 231: Ah so that's what you were explaining this morning. We explicitly tells the other declaration to not install and this one will take care of it. Too bad helm can't resolve this properly like proper dependency managers.

jimleroyer (approved):
LGTM (which means Lucky-Goldstar Trademark, right?)

PR #4805, opened 2026-05-26
Revert "applying goldilocks cpu recommendations core tasks"

whabanks:
Reverts cds-snc/notification-manifests#4804

Celery pods in staging are capped out on CPU and are likely the cause of bulk jobs, and thus the smoke tests, failing. This is blocking [today's release](https://github.com/cds-snc/notification-manifests/pull/4799)


P0NDER0SA (approved):
yup

PR #4894, opened 2026-06-29
preload admin in dev and staging

ben851:

Setting gunicorn preload app to true for the admin in dev and staging. This will greatly reduce the amount of memory needed to load the admin. See below screenshot. The old pods were using 1gb of memory, the new ones with preload on use 200MB.

image

## What are you changing?

Trying to make better use of our k8s nodes.

## After merging this PR


P0NDER0SA (approved):
sure let's give it a try....

PR #4965, opened 2026-07-16
Fix: Reorder setup-cluster-access steps (AWS creds before VPN config)

P0NDER0SA:
## Problem
The `setup-cluster-access` action was attempting to retrieve the VPN configuration before AWS credentials were configured. The `createVPNConfig.sh` script uses `terragrunt output` which requires AWS authentication, causing the workflow to fail when trying to query Terraform state from S3.

## Solution
Reordered the steps so that "Configure AWS credentials" happens before "Retrieve VPN Config". This ensures AWS credentials are available when terragrunt needs them.

## Changes
- Moved "Configure AWS credentials" step before "Retrieve VPN Config" step
- All other steps remain in the same order

## Step Order (Now Correct)
1. Install 1Password CLI
2. Setup Terraform tools
3. **Configure AWS credentials** ← moved up
4. **Retrieve VPN Config** ← moved down
5. Install OpenVPN
6. Connect to VPN
7. Configure kubeconfig

### DIFF
--- a/.github/actions/setup-cluster-access/action.yaml
+++ b/.github/actions/setup-cluster-access/action.yaml
@@ -31,6 +31,13 @@ runs:
       env:
         TERRAGRUNT_VERSION: 0.66.9
 
+    - name: Configure AWS credentials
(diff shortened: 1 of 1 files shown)


ben851 (approved):
lgtm

PR #4966, opened 2026-07-16
Fix: Use correct IAM role names for Velero workflows per environment

P0NDER0SA:
## Problem
The Velero workflows were using an incorrect IAM role naming pattern for all environments. The role `notification-manifests-helmfile-dev-apply` does not exist in AWS.

## Actual AWS Roles
After reviewing existing workflows, the correct IAM roles are:
- **Dev**: `notification-manifests-k8s-lambda-apply-main-branch` (different pattern)
- **Staging**: `notification-manifests-helmfile-staging-apply`
- **Production**: `notification-manifests-helmfile-production-apply`

## Solution
Updated all three Velero workflows to use conditional expressions that map to the correct role name per environment:
- Dev uses the special `notification-manifests-k8s-lambda-apply-main-branch` role
- Staging and Production use the `notification-manifests-helmfile-{env}-apply` pattern

## Changed Files
- `.github/workflows/velero_list_backups.yaml`
- `.github/workflows/velero_snapshot.yaml`
- `.github/workflows/velero_restore.yaml`

This fixes the infinite OIDC assumption loop caused by the non-existent role.


ben851 (approved):
maybe maybe maybe

PR #4980, opened 2026-07-21
Fix `setup-cluster-access` action Terragrunt VPN failures

P0NDER0SA:
## Description
Fixes a pervasive issue where `setup-cluster-access` would panic during the VPN configuration step due to Terragrunt lockfile failures.

The root cause was a mismatch in tooling versions: the action previously failed to pin a specific `TERRAFORM_VERSION`, downloading the newest TF binary, which then poisoned the global plugin cache during an extraneous cache-warming step. When `createVPNConfig.sh` subsequently ran Terragrunt pinned to older checksums from `.terraform.lock.hcl`, the cached plugins threw `does not match any of the checksums recorded in the dependency lock file` errors.

### Changes
* Removed the `Cache Terraform providers` block from `setup-cluster-access` as plugins are securely downloaded by `createVPNConfig.sh` natively anyway.
* Pinned the exact tool versions in `terraform-tools-setup` to mirror our standard `helmfile_staging_apply.yaml` setup (TF 1.9.5, TG 0.66.9, etc.).


ben851 (approved):
lgtm

PR #5005, opened 2026-07-26
TGB migration script improvements

ben851:

Saving after the fact - had to make two adjustments to the migration script for target group bindings:

1. Assume DNS role if the environment is production
2. Halt the script if the first terraform step fails.

## What are you changing?

- [x] Script

IP target group migration

## After merging this PR


jimleroyer (approved):
LGTM

PR #5020, opened 2026-07-28
feat: Removing static fair queue config

jimleroyer:

  - `BREAKING CHANGE:` - tag `main` as a new major release
  - `chore:` - use for changes to non-app code (ex: GitHub actions)
- Alternatively, change the [VERSION file](https://github.com/cds-snc/notification-manifests/blob/main/VERSION) - this will not create a new tag, but rather will release the tag in `VERSION` to production.

## What are you changing?

Removing static fair queue config. We don't need these anymore as existing Celery instances that send SMS will be used, instead of new ones that were tuned to the new AWS fair queue meant for the rate limited SMS, i.e. existing SMS queues will be rate limited. 

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document download API
- [ ] notification-lambdas

  - [ ] notification-lambdas ([heartbeat](https://ca-central-1.console.aws.amazon.com/ecr/repositories/private/296255494825/notify/heartbeat?region=ca-central-1), [ses_to_sqs_email_callbacks](https://ca-central-1.console.aws.amazon.com/ecr/repositories/private/296255494825/notify/ses_to_sqs_email_callbacks?region=ca-central-1), [system-status](https://ca-central-1.console.aws.amazon.com/ecr/repositories/private/296255494825/notify/system_status?region=ca-central-1))

## After merging this PR


P0NDER0SA (approved):
Yep!  good job

helmfile/overrides/notify/celery.yaml.gotmpl
```diff
@@ -251,13 +236,13 @@ nodes:
           memory: {{ if eq .Environment.Name "production" }} "800Mi" {{ else if eq .Environment.Name "staging" }} "800Mi" {{ else }} "800Mi" {{ end }}
         limits:
           cpu: {{ if eq .Environment.Name "production" }} "550m" {{ else if eq .Environment.Name "staging" }} "700m" {{ else }} "550m" {{ end }}
-          memory: {{ if eq .Environment.Name "production" }} "1024Mi" {{ else if eq .Environment.Name "staging" }} "1024Mi" {{ else }} "1024Mi" {{ end }}
+          memory: {{ if eq .Environment.Name "production" }} "1024Mi" {{ else if eq .Environment.Name "staging" }} "1280Mi" {{ else }} "1024Mi" {{ end }}
     autoscaling:
       hpaEnabled: {{ if eq .Environment.Name "production" }} true {{ else if eq .Environment.Name "staging" }} true {{ else }} true {{ end }}
       minReplicas: {{ if eq .Environment.Name "production" }} 2 {{ else if eq .Environment.Name "staging" }} 2 {{ else }} 2 {{ end }}
       maxReplicas: {{ if eq .Environment.Name "production" }} 2 {{ else if eq .Environment.Name "staging" }} 2 {{ else }} 2 {{ end }}
```
ben851:
```suggestion
      maxReplicas: {{ if eq .Environment.Name "production" }} 2 {{ else if eq .Environment.Name "staging" }} 3 {{ else }} 2 {{ end }}
```

helmfile/overrides/notify/celery.yaml.gotmpl
```diff
@@ -251,13 +236,13 @@ nodes:
...
       targetCPUUtilizationPercentage: {{ if eq .Environment.Name "production" }} 25 {{ else if eq .Environment.Name "staging" }} 25 {{ else }} 25 {{ end }}
-      scaleUpPodsValue: {{ if eq .Environment.Name "production" }} 6 {{ else if eq .Environment.Name "staging" }} 6 {{ else }} 6 {{ end }}
+      scaleUpPodsValue: {{ if eq .Environment.Name "production" }} 6 {{ else if eq .Environment.Name "staging" }} 7 {{ else }} 6 {{ end }}
```
ben851, line 245: This line indicates how many "steps" celery should scale up by when autoscaling... So if it recognizes the need to autoscale, it will add up to x number of new pods assuming it doesn't go over max replicas. So this doesn't need to be changed to add more pods.

helmfile/overrides/notify/celery.yaml.gotmpl
```diff
@@ -251,13 +236,13 @@ nodes:
           memory: {{ if eq .Environment.Name "production" }} "800Mi" {{ else if eq .Environment.Name "staging" }} "800Mi" {{ else }} "800Mi" {{ end }}
         limits:
           cpu: {{ if eq .Environment.Name "production" }} "550m" {{ else if eq .Environment.Name "staging" }} "700m" {{ else }} "550m" {{ end }}
-          memory: {{ if eq .Environment.Name "production" }} "1024Mi" {{ else if eq .Environment.Name "staging" }} "1024Mi" {{ else }} "1024Mi" {{ end }}
+          memory: {{ if eq .Environment.Name "production" }} "1024Mi" {{ else if eq .Environment.Name "staging" }} "1280Mi" {{ else }} "1024Mi" {{ end }}
```
ben851: Do we really need more memory for celery?

ben851 (changes requested):

helmfile/overrides/notify/celery.yaml.gotmpl
jimleroyer: Yeah I changed that at first, but then I saw scaleUpPodsValue and wondered, why not that one? I just read your comment below though, thanks.

helmfile/overrides/notify/celery.yaml.gotmpl
jimleroyer: These values seem high given that the current # of replicas are much lower.

helmfile/overrides/notify/celery.yaml.gotmpl
jimleroyer: You tell me, we'll have 1 more replica which can take 600-800 MiB (without the memory improvement). What do you think?

ben851: This is memory per pod, not shared across all pods, so you shouldn't have to change this, it will just be one more pond that will take 600-800mb from the cluster

helmfile/overrides/notify/celery.yaml.gotmpl
ben851: Yes, we originally had several scalable pods but we've effectively disabled the HPA because we've set min and max replicas to 2, making this superfluous.

ben851 (approved):

PR #5081, opened 2026-08-18
chore: pin floating/mutable Helm-deployed container image tags

P0NDER0SA:
## Description

Split out of #5067 to keep that PR focused on GitHub Actions SHA pinning. This PR covers the "Helm charts and Helm references should be covered" acceptance criterion from the supply-chain-hardening card — pinning container image references that Helm charts in this repo deploy, which were previously floating on mutable `:latest` (or unset/implicit `:latest`) tags.

**Merge order:** land after #5067 (independent files, no conflict, but keeping the GH Actions PR as the first/primary supply-chain change).

## Changes

- `helmfile/overrides/system/ipv4-geolocate.yaml.gotmpl`: was empty (chart default `tag: latest` was always deployed with no override). Pinned to `2026-08-12` (latest available tag on `public.ecr.aws/cds-snc/notify-ipv4-geolocate-webservice`).
- `helmfile/overrides/system/velero.yaml.gotmpl`: removed the `kubectl.image` override (`repository: public.ecr.aws/bitnami/kubectl`, `tag: "latest"`). Live registry check confirmed that repository path returns `404 NAME_UNKNOWN` — Bitnami retired that mirror. The vmware-tanzu/velero chart's own default (`bitnamilegacy/kubectl` on Docker Hub, no tag) auto-selects a kubectl version matching the cluster's own Kubernetes version, which is both correct and safer than us hardcoding a version that could drift from the cluster.
- `helmfile/overrides/tools/blazer.yaml.gotmpl`: `migrate` initContainer image pinned to `ankane/blazer:v3.5.1` (was untagged, implicit `:latest`). The main `blazer` deployment image was already pinned (custom ECR tag) and is untouched.
- `helmfile/charts/signoz-importer/values.yaml`: `job.image.tag` pinned via combined `tag@digest` syntax (`latest@sha256:eb00b3d4864c...`) since `dwdraju/alpine-curl-jq` publishes no semver tags upstream, only `latest`.

## QA Steps

- [x] Confirmed via public registry APIs (Docker Hub v2, public ECR anonymous token flow) that each new/kept tag or digest actually exists and resolves.
- [x] Confirmed `public.ecr.aws/bitnami/kubectl` no longer exists (404) before deciding to drop the override rather than re-pin it.
- [x] YAML validity spot-checked with `yq` (and manual review for the two files containing Go-template syntax that `yq` can't parse directly).
- [ ] Deploy to sandbox/dev and confirm `ipv4-geolocate`, `velero` (CRD/cleanup jobs), `blazer` (migrate initContainer), and `signoz-importer` jobs still pull and run successfully with the new image references.

## Not in scope (flagged separately)

- `helmfile/overrides/dev.env`'s literal `"latest"` DOCKER_TAG values — left alone, likely intentionally managed by `update_image_tags_dev.yaml`.
- `helmfile/overrides/system/github-arc-scaleset.yaml.gotmpl`'s `default "latest"` fallback — the env vars that would override it aren't defined anywhere in this repo, so behavior couldn't be confirmed.


helmfile/overrides/system/ipv4-geolocate.yaml.gotmpl
ben851: Have you validated this image exists? I'm thinking this should take the same approach as we do with api/admin for image tags

ben851 (approved):
LGTM with the one note

PR #4814, opened 2026-05-28
NodeSelector Tweak for DB Upgrade K8s Job

P0NDER0SA:
This pull request updates the configuration for the `notify-database` Helm chart to improve node scheduling flexibility and support for tolerations. The main changes involve switching from on-demand to spot instances and adding support for specifying tolerations, both in the chart values and the job template.

**Node scheduling updates:**

- Changed the node selector label from `eks.amazonaws.com/capacityType: ON_DEMAND` to `karpenter.sh/capacity-type: spot` in both `values.yaml` and the override file, switching the default scheduling from on-demand to spot instances. [[1]](diffhunk://#diff-2f631009629feff01884cc4b9a4e4b5aa72bb389dfdf9f6b0bc62b5dafb7d3f3L23-R25) [[2]](diffhunk://#diff-82b49a58116716f45a784a218e758b9fb3cf6453b4ca62636af336f885cedec3L103-R117)

**Tolerations support:**

- Added a `tolerations` field to `values.yaml` (defaulting to an empty list) to allow specifying tolerations for the database job.
- Updated the job template (`run-database-job.yaml`) to include tolerations if they are defined in values.
- Set specific tolerations in the override file to allow the job to schedule on nodes with certain taints related to monitoring and storage agents.

## What are you changing?

- [ ] Releasing a new version of Notify
- [ ] Changing kubernetes configuration

## Provide some background on the changes

> Give details ex. Security patching, content update, more API pods etc

## If you are releasing a new version of Notify, what components are you updating

- [ ] API
- [ ] Admin
- [ ] Documentation
- [ ] Document download API
- [ ] notification-lambdas

## Checklist if releasing new version

- [ ] I made sure that the changes are as expected in [Notify staging](https://staging.notification.cdssandbox.xyz/)
- [ ] I have checked if the docker images I am referencing exist
  - [ ] [api lambda](https://ca-central-1.console.aws.amazon.com/ecr/repositories/private/296255494825/notify/api-lambda?region=ca-central-1) (requires Notification-Production / AdministratorAccess login)
  - [ ] [api k8s](https://gallery.ecr.aws/v6b8u5o6/notify-api)
  - [ ] [admin](https://gallery.ecr.aws/v6b8u5o6/notify-admin)
  - [ ] [documentation](https://gallery.ecr.aws/v6b8u5o6/notify-documentation)
  - [ ] [document download API](https://gallery.ecr.aws/v6b8u5o6/notify-document-download-api)
[...]

PR #5101, opened 2026-08-19
Fix intermittent DB migration Job timeouts by using on-demand nodes

P0NDER0SA:
## Problem

The `notify-db-migration-job` (run by every deploy to dev/staging/production via `.github/actions/db-migration`) intermittently times out. The wait step (`kubectl wait --for=condition=complete ... --timeout=400s`) has ~6.7 minutes total, but the job's pod is pinned to Karpenter spot nodes only:

```yaml
nodeSelector:
  karpenter.sh/capacity-type: spot
```

When there's no already-warm spot node with room, Karpenter has to provision a brand new one (EC2 spot capacity fulfillment + node bootstrap), which can eat several minutes on its own, leaving little time for the actual migration before the GitHub Action's 400s wait gives up - blocking staging and production releases.

## History / why this is safe

Checked git history for this field before touching it:
- The job originally (Mar 2025, #3652) used `eks.amazonaws.com/capacityType: ON_DEMAND` - the same always-warm EKS managed on-demand node group already used today by `notify-api`, `notify-admin`, `notify-celery`, `notify-document-download`, and most system components.
- PR #4814 switched it to a Karpenter-style selector, which didn't actually work (this cluster's Karpenter NodePool only provisions `spot` - there's no Karpenter on-demand pool), so PR #4815 patched it to `spot` the very next day as a stopgap ("spot because that's all karpenter allows").
- Nobody went back to the original on-demand config - this PR is effectively reverting to that known-good state, not introducing something new/untested.

## Fix

1. `helmfile/charts/notify-database/values.yaml`: nodeSelector back to `eks.amazonaws.com/capacityType: ON_DEMAND`.
2. `helmfile/overrides/notify/database.yaml.gotmpl`: this release's only values file *also* set `nodeSelector: karpenter.sh/capacity-type: spot` plus tolerations for nidhogg startup taints, which would have been deep-merged on top of the chart's on-demand default - producing an impossible nodeSelector requiring both labels on the same node (permanently unschedulable, worse than today). Removed that stale override so the chart default applies cleanly.

## Verification

- `helm lint`/`helm template` on the chart alone confirms the rendered Job requests `eks.amazonaws.com/capacityType: ON_DEMAND`.
- Reproduced the real merge with a `helm template -f <override values>` test: confirmed the *before* state produces a broken two-key nodeSelector, and the *after* state (override removed) produces a single correct `eks.amazonaws.com/capacityType: ON_DEMAND` selector.

### DIFF
--- a/helmfile/charts/notify-database/values.yaml
+++ b/helmfile/charts/notify-database/values.yaml
@@ -19,8 +19,12 @@ serviceAccount:
   create: true
   serviceAccountName: "notify-database"
 
+# Run on the always-warm EKS managed on-demand node group instead of a Karpenter
+# spot node: a new spot node can take 5+ minutes to provision and become
+# schedulable (spot capacity fulfillment + node bootstrap + nidhogg startup
+# taints), which blocks this job and any deploy waiting on it.
 nodeSelector:
-  karpenter.sh/capacity-type: spot
+  eks.amazonaws.com/capacityType: ON_DEMAND
 
 tolerations: []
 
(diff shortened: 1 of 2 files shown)


ben851 (approved):
lgtm

PR #5107, opened 2026-08-20
feat: resolve dev image tags from latest pushed build instead of :latest

P0NDER0SA:
## What
Instead of resolving dev's image tags from the most-recently-pushed ECR image on a weekly cron, dev now simply mirrors whatever tag staging was just updated to, in the same commit.

- `update_image_tags_staging.yaml` (renamed `update_image_tags_staging_and_dev.yaml`) now writes the component's tag to **both** `helmfile/overrides/staging.env` and `helmfile/overrides/dev.env` in a single commit, whenever the existing `update-docker-image` repository_dispatch fires.
- Removed `update_dev_image_tags_weekly.yaml` (the ECR-tag-resolving cron workflow) - no longer needed.
- Removed `update_image_tags_dev.yaml` (the separate `update-docker-image-dev` dispatch workflow) - dev is now driven entirely by the staging update, so this dispatch type is unused going forward.
- One-time fix already applied: `dev.env`'s `ADMIN/API/DOCUMENT_DOWNLOAD/DOCUMENTATION_DOCKER_TAG` were hardcoded to `"latest"`; they now hold real resolved tags and will be kept in sync automatically from here on.

## Why
Simpler than resolving tags out-of-band: dev always reflects "whatever staging is currently running," updated at the same time staging is, with no extra scheduled job or ECR API calls to maintain.

## Follow-up (not in this PR)
Several other pieces of the broader "no latest/bootstrap tags" effort still default to mutable tags when the override isn't wired through (e.g. ipv4-geolocate chart default, some Terraform lambda fallbacks in notification-terraform, and CI in a few app repos that still push `:latest` as a side effect). Tracked separately.


jimleroyer (approved):
LGTM

